PR #17, opened 2025-03-16
Add i18n and Chinese language

alanpeng:
Add i18n and Chinese language

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,10 +143,12 @@ MCP connects AI models to external tools like web browsers, documentation system
 
 ### Key Features:
 
-- **Tool Integration**: Connect to browsers, documentation systems, and language-specific tools
-- **Provider Options**: Choose between OpenAI and Anthropic models
-- **Flexible Configuration**: Enable/disable servers, customize settings, and control usage
-- **Seamless Workflow**: MCP tools gather information, then Aider implements the code changes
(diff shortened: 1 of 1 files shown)


wladimiiir:
Thank you for your interest and contribution. 

I can see several issues that need to be addressed to get this merged into the main:
1. to have the proper translation foundation in place, we need to use `react-i18next` library and utilize `useTranslation` hook.
2. translation files need to have proper key organized by their context - for example `t("Min Time Between Tool Calls (ms)")` needd to be `t("settings.mcpSettings.minTimeBetweenToolCalls")`
3. please replace Chinese comments with English ones
4. there are some other smaller issues that I would comment later, when you decide to work on this

But seeing that multi-language support is something requested, I might work on this next week to prepare the foundation and then I would just kindly ask you to prepare zh.json translation file from the prepared en.json.

Let me know if you would like to work on this further, or you would rather have me take over it. Thanks.

wladimiiir:
@alanpeng 
If you are interested have a look at [support-other-languages](https://github.com/hotovo/aider-desk/tree/support-other-languages) branch. The foundation is there, what needs to be done is to replace all the strings with `t(...)` and add the translations in the `en.ts`, `zh.ts` files.
Thanks.

PR #25, opened 2025-03-29
Support other languages

alanpeng:
support i18n and add Chinese language traslation


alanpeng:
image
image
image
image
image

src/renderer/src/pages/Onboarding.tsx
```diff
@@ -1,22 +1,41 @@
...
+        console.log('Finishing setup: Saving settings with onboardingFinished=true');
+        // Save the settings first, which will trigger the useEffect in LanguageSwitcher to apply the language change
+        await saveSettings({
+          ...settings!,
+          onboardingFinished: true,
+        });
+        console.log('Settings saved successfully, navigating to Home');
+        
+        // Use window.location.hash for navigation in HashRouter context
+        console.log('Using window.location.hash for navigation in Electron with HashRouter');
```
wladimiiir, line 28: Why is there a need for this change instead of using `navigate`?

wladimiiir:
Hi @alanpeng,
thanks again for your contribution. I have made the required changes in `main` branch and added support for the Chinese language. I am myself not really able to test the language properly, as the characters do not show up correctly for me, so I would like to kindly ask you if you could have a look and test it. In case of any issues or changes needed, please create a new PR with the fixes.

As for this PR, we should be able to close it.  Thanks a lot again.

Issue #79, opened 2025-04-30
Show detailed output from /tokens


greinacker:
I like the tokens and cost box in the lower right of the window; however, one of my favorite Aider features is being able to type /tokens, and see exactly what I have in my context, and the breakdown of how many tokens each item takes up.

So, I suggest allowing the user to type /tokens, and show the output, something like:

<img width="1416" alt="Image" src="https://github.com/user-attachments/assets/286482e2-ce5d-4b8b-be95-689e6466bf82" />

wladimiiir:
Thanks for the feedback.

This command was intentionally left out because I added the information to the cost info section (bottom right). I also planned to display token information for each file on hover, but haven’t had the chance to implement that yet.

That said, I’m currently working on improving the cost section itself, as it’s still missing proper token calculations in Agent mode.

But adding a `/tokens` command as addition to this won't hurt, so let's go with it.

Thanks.

greinacker:
Ya I was also thinking having it shown in the file section would be cool at some point. But, since everyone who uses aider knows about /tokens, this will be nice to have in the meantime!

wladimiiir:
This has been implemented in 0.11.1. Thanks for your contribution.

PR #80, opened 2025-04-30
add /tokens command

greinacker:
Adds the /tokens command, and displays the output to the user.

Implements #79

### DIFF
--- a/resources/connector/connector.py
+++ b/resources/connector/connector.py
@@ -657,6 +657,9 @@ class Connector:
     if command.startswith("/test ") or command.startswith("/run "):
       self.coder.io.running_shell_command = True
       self.coder.io.tool_output("Running " + command.split(" ", 1)[1])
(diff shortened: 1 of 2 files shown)


resources/connector/connector.py
```diff
@@ -657,6 +657,9 @@ async def run_command(self, command):
     if command.startswith("/test ") or command.startswith("/run "):
       self.coder.io.running_shell_command = True
       self.coder.io.tool_output("Running " + command.split(" ", 1)[1])
+    elif command.startswith("/tokens"):
+      self.coder.io.running_shell_command = True
```
wladimiiir, line 661: Not ideal to have it as a command output here, but given Aider’s `cmd_tokens` implementation, using `send_log_message` would require copying too much code into the connector. So let’s stick with this solution.

wladimiiir (approved):
Thanks :+1:

Issue #85, opened 2025-05-01
Enable all commands


pcfreak30:
Every command aider supports, desk should support without additional sync ideally in updates? The most obvious now is the `/drop` cmd.

wladimiiir:
Thanks for the feedback.

The `/drop` command will be added in #65.

Are there any other specific commands you'd like to see supported? While not all commands are included by default - since some require special handling in AiderDesk - the most common ones have been prioritized, and others can be added based on demand and feasibility.

Issue #84, opened 2025-05-01
Adding/removing context is extremely slow


pcfreak30:
I am not sure if this is something on my end, but if i have like 15 files, and im closing out everything in idea, it is dialup slow in removing all the files.

Additionally I second https://github.com/hotovo/aider-desk/issues/65 that we need to be able to more easily drop.

wladimiiir:
Thanks for the feedback.

Could you share a short video showing the behavior? IntelliJ IDEA currently sends 'drop' events one by one, which might explain why the files appear to disappear sequentially, but this should not be 'extremely slow'.

pcfreak30:
https://github.com/user-attachments/assets/5376590f-0e6b-4699-88bd-7cf1ee7c9105

You may be right on the IDEA side. but its also something with aider that it does not handle a huge file context list. it also is similar in CLI and i can't find any issues on it, though I thought one existed.

The UX on this is really bad. There should also be a button to drop all files too.

wladimiiir:
Thanks for the follow-up.

1. How many files do you usually have in the context?
2. When you said "it also is similar in CLI", can you confirm if you're seeing the same slowness or lag when using Aider directly from the command line, without AiderDesk?

I'd like to narrow down whether the performance bottleneck is coming from AiderDesk's interface or from Aider's internal processing.

pcfreak30:
> 2\. When you said "it also is similar in CLI", can you confirm if you're seeing the same slowness or lag when using Aider directly from the command line, without AiderDesk?

Yes, Aider tend to have issues having 50+ files in its context and I would honestly love to solve THAT overall.

wladimiiir:
I was able to identify the issue and improve the management logic. For reference, the slowdown was caused by the Autocompleter tokenizing files—this process can take several seconds when dealing with a large number of files. Aider performs this internally when using Aider, but since AiderDesk handles it within the connector, I was able to optimize it.

@pcfreak30
Are you using the released version, or do you also have a development setup? If you do, feel free to test it out and let me know your experience. Thanks.

pcfreak30:
> I was able to identify the issue and improve the management logic. For reference, the slowdown was caused by the Autocompleter tokenizing files—this process can take several seconds when dealing with a large number of files. Aider performs this internally when using Aider, but since AiderDesk handles it within the connector, I was able to optimize it.
> 
[...]

Issue #65, opened 2025-04-28
Feature : Drop files using slash drop command


baramofme:
The user should be able to remove added context files from the current session using the drop slash command.

wladimiiir:
Thanks for the feedback.

Do you mean `/drop` for dropping all files? Not sure if removing them individually would make sense.

baramofme:
Yes, Drop everyfile make sense. sometimes added files too many.

wladimiiir:
Support for `/drop` command was added in [0.12.0](https://github.com/hotovo/aider-desk/releases/tag/v0.12.0).

PR #97, opened 2025-05-03
implement reset command

greinacker:
Not so sure about this, but I wanted to at least send something concrete for discussion.

There have been several issues (#65 and #85 ) posted about being able to drop files, and specifically about dropping _all_ files. So it seems like implementing /reset would be a straightforward way to accomplish this.

So the idea in this PR to implement the /reset command, to a) clear the messages, and b) drop all files from context.

Two things - 

1. The code does indeed work, and updates the file list and the new token bar in the lower right, but it somehow "feels" gross, like we're adding special-case code to somewhere it doesn't belong. That said, better options weren't entirely obvious, so I thought you might be able to either a) agree this is the way to do it, or b) point me in a better direction.

2. As has been mentioned (#84 ), dropping a lot of files at once in Aider seems to be _very_ slow. I tried adding most of the files in the aider-desk repo, then reset, and the Aider python process went to 100% for like a minute or so getting this done. It did work in the end - but there wasn't feedback about it. Not sure that's really our problem here, but I wanted to mention it...

Thanks!

### DIFF
--- a/resources/connector/connector.py
+++ b/resources/connector/connector.py
@@ -686,6 +686,10 @@ class Connector:
         self.thinking_tokens = None
       await asyncio.sleep(0.1)
       await self.send_current_models()
+    elif command.startswith("/reset"):
+      await self.send_update_context_files()
+      await self.send_autocompletion()
+      await self.send_tokens_info()
 
   async def send_autocompletion(self):
     try:
--- a/src/main/project.ts
+++ b/src/main/project.ts
@@ -623,6 +623,12 @@ export class Project {
     }
 
     logger.info('Running command:', { command });
+
+    if (command.trim() === 'reset') {
+      this.sessionManager.clearMessages();
+      this.mainWindow.webContents.send('clear-project', this.baseDir, true, false);
+    }
+
     if (addToHistory) {
       void this.addToInputHistory(`/${command}`);
     }
--- a/src/renderer/src/components/PromptField.tsx
+++ b/src/renderer/src/components/PromptField.tsx
@@ -12,7 +12,20 @@ import { showErrorNotification } from '@/utils/notifications';
 import { ModeSelector } from '@/components/ModeSelector';
 
 const COMMANDS = ['/code', '/context', '/agent', '/ask', '/architect', '/add', '/model', '/read-only'];
-const CONFIRM_COMMANDS = ['/clear', '/web', '/undo', '/test', '/map-refresh', '/map', '/run', '/reasoning-effort', '/think-tokens', '/copy-context', '/tokens'];
+const CONFIRM_COMMANDS = [
+  '/clear',
+  '/web',
+  '/undo',
+  '/test',
+  '/map-refresh',
+  '/map',
+  '/run',
+  '/reasoning-effort',
+  '/think-tokens',
+  '/copy-context',
+  '/tokens',
+  '/reset',
+];
 
 const ANSWERS = ['y', 'n', 'a', 'd'];
 


pcfreak30:
```dropping a lot of files at once in Aider seems to be very slow. I tried adding most of the files in the aider-desk repo, then reset, and the Aider python process went to 100% for like a minute or so getting this done. It did work in the end - but there wasn't feedback about it. Not sure that's really our problem here, but I wanted to mention it...```

Is something that really does need to get solved and I might even try to spend a bit of time on it if your test case is reproducible. some days aider adding/removing files is fast, other times its dog slow.

wladimiiir:
Thanks for your interest in this! For now, I believe we only need to add the `/drop` command. A special `/reset` command that simply combines `/clear` and `/drop` might not be necessary at this point.

Regarding the performance issues, I’m planning to take a closer look at #84 tomorrow to see if there’s anything that can be improved.

greinacker:
Hmm ok. I mean, /reset isn't really special, per se - it's a standard aider command.

But I didn't realize that in aider, /drop without any filename listed would drop all files. That we could do.

For the normal /drop behavior, where aider takes a filename (with autocomplete), it's unclear how that should work here. /add is the closest analogue (as it also uses autocomplete in aider), but here we pop up a dialog...so hard to say how /drop for a single file should be implemented.

If you're thinking just implement /drop with no arguments to drop all files, that makes sense. I think it's pretty much the same implementation I show here, of course swapping out /drop for /reset in PromptField.tsx.

Would you like me to make that change? Or are you thinking you'll do it when you look at #84?

greinacker:
You know, playing around in Aider, if I do (in this repo):

```/add src/*```

It will add 135K tokens or so into the context. If I then do

```/drop```

It drops them all, almost immediately. I repeated this a few times, with no delay.

So I'm thinking maybe it's only slow when dropping files one at a time.

If that's the case, and we want to implement drop all, I think we could just rearrange the code there in project.ts (in this PR's change), to do the runCommand first (which will run /drop on aider), then drop the files from the "local" context, and probably have near-instant performance.

pcfreak30:
> You know, playing around in Aider, if I do (in this repo):
> 
> `/add src/*`
> 
> It will add 135K tokens or so into the context. If I then do
> 
> `/drop`
> 
> It drops them all, almost immediately. I repeated this a few times, with no delay.
> 
> So I'm thinking maybe it's only slow when dropping files one at a time.
> 
> If that's the case, and we want to implement drop all, I think we could just rearrange the code there in project.ts (in this PR's change), to do the runCommand first (which will run /drop on aider), then drop the files from the "local" context, and probably have near-instant performance.

There is some cases though where aider just slows to a craw with context control. I don't remember if its with add/remove or with the cli feeling like im back in AOL...

wladimiiir:
@greinacker 

> Hmm ok. I mean, /reset isn't really special, per se – it's a standard aider command.

You're absolutely right - my mistake. Let's go ahead and keep this PR focused on the `/reset` command.

To maintain the best performance, here’s what I suggest the command should do:

1. `this.sessionManager.clearMessages();`
2. `this.mainWindow.webContents.send('clear-project', this.baseDir, true, false);`
3. Call the `reset` command on Aider
4. In `connector.py`, within `run_command`, for `/reset`, add:

   ```python
   await self.send_update_context_files()
   await self.send_autocompletion()
   await self.send_tokens_info()
   ```

I might be missing a minor detail or two, but this should give you a solid framework to start with.

Thanks.

wladimiiir:
@pcfreak30 
> There is some cases though where aider just slows to a craw with context control. I don't remember if its with add/remove or with the cli feeling like im back in AOL...

In general, I don’t think Aider was designed to handle a large number of files in context. Ideally, your context should include only the files you expect to actively edit during a session. For broader reference, it might be better to consolidate everything into a large llms.txt file (similar to repomix) and include that as read-only.

That said, I’ll look into how the number of context files affects performance - but if Aider itself struggles with this, it’s unlikely that AiderDesk will be able to fully resolve it either.

pcfreak30:
> @pcfreak30
> 
> > There is some cases though where aider just slows to a craw with context control. I don't remember if its with add/remove or with the cli feeling like im back in AOL...
> 
> In general, I don’t think Aider was designed to handle a large number of files in context. Ideally, your context should include only the files you expect to actively edit during a session. For broader reference, it might be better to consolidate everything into a large llms.txt file (similar to repomix) and include that as read-only.
> 
> That said, I’ll look into how the number of context files affects performance - but if Aider itself struggles with this, it’s unlikely that AiderDesk will be able to fully resolve it either.

My view is a PR to aider should be made to solve it.

greinacker:
> I might be missing a minor detail or two, but this should give you a solid framework to start with.

Thank you! Updated as suggested, seems to work as expected.

Having the /tokens command implemented (from the merged PR a few days ago) is super handy for testing this kind of thing! :)

wladimiiir (approved):
Thanks :+1:

Issue #38, opened 2025-04-14
Cannot select local models available in Ollama


CreativeWarlock:
**Problem:**

After 
- selecting 'Ollama' as the provider,
- assigning the URL to the server (http://localhost:11434) and
- selectinig a model that is pulled locally (e.g. deepseek-r1:8b or gemma3:12b)

Aider-Desk does not update the dropdown list with the configured local models.

![Image](https://github.com/user-attachments/assets/9cf73ca8-b433-4633-917e-2a529406c117)

**Suggestion:**

It would be very great if the configuration logic in Aider-Desk dynamically reads the list of models available to Ollama (as in 'ollama list'?) after the user selected the 'Ollama' provider and then place the findings in the model dropdown.

![Image](https://github.com/user-attachments/assets/f23aa4dc-c67e-4d6a-82bd-adaadaf2ebfc)

wladimiiir:
Hi, thanks for the feedback.
I think that part of configuration is probably confusing and I plan to update it.

There are 2 independent model configurations here:
- agent has its own model config via the settings dialog
- the models you see at the top are only for Aider and are used when Aider works on the code

Hope that makes sense.

CreativeWarlock:
So, apologies if I'm harassing you with the desire to get a local LLM connected to the Aider-Desk again.

Your explanations make sense, however, I'm not able to query anything to the local LLM (deepseek-r1:8b) selected through Ollama. Could you assist with the steps how to make it work - or is this an open task that needs implementing?

What model am I supposed to select? There's deepseek-chat and deepseek/deepseek-chat and there are various ollama/llama3:* options offered as well.
![Image](https://github.com/user-attachments/assets/fb49db05-87f3-4a20-9633-89dd92c5d0b8)

wladimiiir:
Hmm, I’ll need to test `deepseek-r1`, but I’m afraid not all models are compatible with Agent mode. It depends on whether the model supports tool usage — and from what I can see, Deepseek R1 doesn’t appear in the [supported models list](https://ollama.com/search?c=tools).

For example, I’m able to use `phi4-mini` in Agent mode without issues.

To set up Ollama for Aider (both the Main and Weak models), you can follow the instructions [here](https://aider.chat/docs/llms/ollama.html).

Here’s an example of a working setup:

![Image](https://github.com/user-attachments/assets/7e0c950c-933d-4b40-905d-1e34da707c68)

CreativeWarlock:
[...]

PR #163, opened 2025-05-21
List available ollama models

greinacker:
In the Settings, Agent tab, this shows a drop-down of all of the currently available models in Ollama. Screenshot:

image

This implements #38

### DIFF
--- a/src/main/ipc-handlers.ts
+++ b/src/main/ipc-handlers.ts
@@ -257,7 +257,10 @@ export const setupIpcHandlers = (
 
   ipcMain.handle('get-ollama-models', async (_, baseUrl: string) => {
     try {
-      const normalized = baseUrl.endsWith('/api') ? baseUrl : `${baseUrl.replace(/\/$/, '')}/api`;
+      let normalized = baseUrl.replace(/\/+$/, ''); // Remove all trailing slashes
+      if (!normalized.endsWith('/api')) {
+        normalized = `${normalized}/api`;
+      }
       const response = await axios.get(`${normalized}/tags`);
       return response.data?.models?.map((m: { name: string }) => m.name) || [];
     } catch (error) {
--- a/src/main/ipc-handlers.ts
+++ b/src/main/ipc-handlers.ts
@@ -2,7 +2,6 @@ import { EditFormat, FileEdit, McpServerConfig, MessageRole, Mode, OS, ProjectDa
 import { normalizeBaseDir } from '@common/utils';
 import { BrowserWindow, dialog, ipcMain } from 'electron';
 import { McpManager } from 'src/main/agent/mcp-manager';
-import axios from 'axios';
 
 import { Agent } from './agent';
 import { getFilePathSuggestions, isProjectPath, isValidPath } from './file-system';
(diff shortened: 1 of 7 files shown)


src/main/ipc-handlers.ts
```diff
@@ -254,6 +255,17 @@ export const setupIpcHandlers = (
     return await mcpManager.getMcpServerTools(serverName, config);
   });
 
+  ipcMain.handle('get-ollama-models', async (_, baseUrl: string) => {
+    try {
+      const normalized = baseUrl.endsWith('/api') ? baseUrl : `${baseUrl.replace(/\/$/, '')}/api`;
```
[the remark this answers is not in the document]
greinacker: I mean, I could change this, but the default URL that's already entered is http://localhost: 11434/api and does not have a trailing slash, so I think it's somewhat unnecessary. That said, just lemme know if you'd prefer I update it.

src/renderer/src/components/settings/providers/OllamaParameters.tsx
```diff
@@ -13,13 +14,28 @@ type Props = {
...
+  const [models, setModels] = useState<string[]>([]);
+
+  useEffect(() => {
+    const loadModels = async () => {
+      if (!baseUrl) {
+        setModels([]);
+        return;
+      }
+      try {
+        const fetched = await window.api.getOllamaModels(baseUrl);
```
wladimiiir: To keep the functionality isolated, I'd prefer `fetch`/`axios` to be used directly within this component (within renderer), as there's no significant need for an IPC call here. Additionally, this could be refactored into a `useOllamaModels` hook that returns the list of models and that could be reused e.g. in `AgentModelSelector`.

src/main/ipc-handlers.ts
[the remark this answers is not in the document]
wladimiiir: I think Gemini made a good point here. The normalization approach expects there could be trailing `/`, so `/` should be removed before the `/api` check.

src/renderer/src/components/settings/providers/OllamaParameters.tsx
greinacker: I might be misunderstanding (likely!), but when I move the retrieval code into a hook that's called by the client, I get an error like the following:

```axios.js?v=6f4b3038:1637 Refused to connect to 'http://localhost:11434/api/tags' because it violates the following Content Security Policy directive: "default-src 'self'". Note that 'connect-src' was not explicitly set, so 'default-src' is used as a fallback.```

One way to fix this is to change the policy in index.html and add **connect-src 'self' http://localhost:***, but that's pretty gross, and wouldn't solve a similar case for Ollama running on a different machine.

So given all that, seems like it makes more sense to have the API calls made from the server, and do it via the IPC call?

src/main/ipc-handlers.ts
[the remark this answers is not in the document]
greinacker: Ha - ok, two against one. :)  Updated!

src/renderer/src/components/settings/providers/OllamaParameters.tsx
wladimiiir: That's a good point. I still believe it's more efficient to keep this in `index.html` rather than creating an additional, specific IPC function. The `connect-src 'self' http: https:` directive should cover all URLs, and this shouldn't pose a problem for a local application like AiderDesk.

Alternatively, we could implement a more general `window.api.fetch` function that handles any URL, method, and body combination, but having `getOllamaModels` seems a bit too specific.

Gemini?

wladimiiir: Yeah, we are not really concerned with the renderer's ability to connect to everything, so CSP should be sufficient.

@greinacker, what are your thoughts?

greinacker: @wladimiiir works for me - I'll prepare an update with relaxed CSP.

greinacker:
Okey dokey...refactored as requested, with the retrieval now being done in the renderer.

Issue #77, opened 2025-04-30
Poor Visibility in Low-Brightness Environments Due to Lack of Theming Options


NADOOITChristophBa:
**Description:**  
The current UI lacks proper theming support, making it difficult to read—especially on low-brightness screens. While scaling adjustments are available, the absence of customizable themes (e.g., dark/light/high-contrast) creates accessibility issues.  

**Specific Problem:**  
In the **Context Files** section, folder names are nearly indistinguishable due to insufficient contrast (too dark). This forces users to rely on guesswork or increase screen brightness, which is not always practical.  

**Suggested Solution:**  
Introduce theming options (e.g., adjustable colors, predefined themes) to improve readability across different environments. At minimum, a higher-contrast mode or folder visibility tweaks would resolve the immediate pain point.  

**Impact:**  
- Hinders usability in low-light conditions.  
- Reduces accessibility for users with visual sensitivities.

wladimiiir:
Thanks for the suggestion - it's definitely something on the backlog.

wladimiiir:
Light mode theme has been added. There is still room for improvement, but that will require more effort and will be done as part of adding support for theming.

PR #164, opened 2025-05-22
Implement light mode theme

greinacker:
So we didn't discuss this beforehand, and normally I'd just make a screenshot and add to the existing issue (#77 from a while back), but I ended up with a fairly simple implementation so I thought I'd put it here in a PR for discussion.

My goal here was to implement an alternate theme, in a light mode, with a bit higher contrast than the default dark mode. Dark mode is unchanged.

Thoughts?

Here are some screenshots:

image

image

image

image

### DIFF
--- a/src/common/types.ts
+++ b/src/common/types.ts
@@ -168,7 +168,7 @@ export interface SettingsData {
   startupMode?: StartupMode;
   zoomLevel?: number;
   notificationsEnabled?: boolean;
-  darkMode?: boolean;
+  theme?: 'dark' | 'light';
   aiderDeskAutoUpdate: boolean;
   aider: {
     options: string;
--- a/src/common/types.ts
+++ b/src/common/types.ts
@@ -168,6 +168,7 @@ export interface SettingsData {
   startupMode?: StartupMode;
   zoomLevel?: number;
   notificationsEnabled?: boolean;
+  darkMode?: boolean;
   aiderDeskAutoUpdate: boolean;
   aider: {
     options: string;
--- a/src/main/store/store.ts
+++ b/src/main/store/store.ts
@@ -21,7 +21,7 @@ export const DEFAULT_SETTINGS: SettingsData = {
   startupMode: StartupMode.Empty,
   zoomLevel: 1,
   notificationsEnabled: false,
-  darkMode: true,
+  theme: 'dark',
   aiderDeskAutoUpdate: true,
   aider: {
     options: '',
--- a/src/main/store/store.ts
+++ b/src/main/store/store.ts
@@ -21,6 +21,7 @@ export const DEFAULT_SETTINGS: SettingsData = {
   startupMode: StartupMode.Empty,
   zoomLevel: 1,
   notificationsEnabled: false,
+  darkMode: true,
   aiderDeskAutoUpdate: true,
   aider: {
     options: '',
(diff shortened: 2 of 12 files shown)


greinacker:
Hey @wladimiiir any thoughts on this? I can close this if you want to table it indefinitely, but I felt like adding a simple light mode would be a good baby step towards multiple themes...

wladimiiir:
@greinacker
I appreciate your effort, but I'm not satisfied with the current screenshots. I understand that the color classes used withing the componets are a mess, and I'm concerned that this "half-baked" solution might require a more comprehensive overhaul. I'd like to experiment with this further when I have more time to determine a better approach.
If you want to experiment with it, perhaps try incorporating some bright, non-grayscale colors for the light theme, such as `#e2e8f0` as the brightest shade, so it's not all so whitey.

wladimiiir:
I'm also curious about the appearance of the diff viewer, assuming it's not affected by light mode.

greinacker:
Ha - I thought the "whitey"-ness looked pretty good. 😄  At the moment this line in main.css controls the overall color hue if you want to play around with it:

```
body.light-mode #root {
  filter: invert(1) hue-rotate(180deg);
}
```

Diffs look pretty good imo:

image

Since this is all super subjective, it feels like one of those things we could go back and forth on 50 times, which probably isn't a great use of either of our time. So, maybe we should just shelf this for the moment, and close this PR?

src/common/types.ts
```diff
@@ -168,6 +168,7 @@ export interface SettingsData {
   startupMode?: StartupMode;
   zoomLevel?: number;
   notificationsEnabled?: boolean;
+  darkMode?: boolean;
```
wladimiiir: Let's have this as `theme: 'dark' | 'light'` instead of a boolean flag, so we can support more themes in the future.

src/main/store/store.ts
```diff
@@ -21,6 +21,7 @@ export const DEFAULT_SETTINGS: SettingsData = {
   startupMode: StartupMode.Empty,
   zoomLevel: 1,
   notificationsEnabled: false,
+  darkMode: true,
```
wladimiiir: `theme: 'dark'`

src/renderer/src/App.tsx
```diff
@@ -11,6 +11,18 @@ import 'react-toastify/dist/ReactToastify.css';
 import { ROUTES } from '@/utils/routes';
 import '@/i18n';
 
+const ThemeManager = () => {
+  const { settings } = useSettings();
+
+  useEffect(() => {
+    const className = settings?.darkMode === false ? 'light-mode' : 'dark-mode';
```
wladimiiir: Let's have `theme-dark` and `theme-light` values.

wladimiiir:
@greinacker
Actually, looking at the screenshots again, the design looks pretty good. It's probably just my preference of dark modes everywhere :slightly_smiling_face:

I've left a few comments.

greinacker:
> @greinacker Actually, looking at the screenshots again, the design looks pretty good. It's probably just my preference of dark modes everywhere 🙂
> 
> I've left a few comments.

Ok! I made the suggested changes. Also made a couple of tweaks to the context file tree, to make folders be dimmed, as they are in the dark theme. So now, only selected files are rendered in the dark color, as expected. New and improved screenshot:

image

wladimiiir (approved):
Awesome job, thanks.

PR #165, opened 2025-05-23
Minor bug fix - update the listener being removed

greinacker:
Fixes a minor bug where input history listeners were removed from the wrong IPC event - looks like a copy/paste typo.

Updated removeInputHistoryUpdatedListener in src/preload/index.ts to detach the listener from the correct "input-history-updated" event


wladimiiir (approved):
Good catch! Thank you.

PR #173, opened 2025-05-27
Made the system python detection smarter

wolph:
Instead of asking the user to fix a python version, try to find a better matching one automatically if they've got multiple versions installed :)

### DIFF
--- a/src/main/start-up.ts
+++ b/src/main/start-up.ts
@@ -45,18 +45,8 @@ export const getOSPythonExecutable = async (): Promise<string> => {
 
     for (const candidate of candidates) {
       try {
-        // For "py -3.12", use shell:true, otherwise append --version
-        let command: string;
-        let useShell = false;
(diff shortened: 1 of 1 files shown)


wladimiiir:
Thank you for your contribution.

During the initial review, I noted that the `shell` property is currently set as a boolean, but it should be a string representing the shell executable (e.g., `cmd.exe`). This will require testing to confirm functionality.

Additionally, could you please confirm if the functionality has been tested across all supported operating systems (Linux, macOS, and Windows)?

Thank you.

wolph:
You're right... that was suggested by copilot while modifying the code but that does look wrong. Having that said, I did test it on windows, linux and os x without any issues but I'll remove that bit just to be sure.

wladimiiir (approved):

wladimiiir:
Thanks for your contribution.

PR #178, opened 2025-06-01
fix: prevent cmd window popups on Windows

FallDownTheSystem:
## Problem

  On Windows, terminal windows briefly appear and steal focus periodically when
  AiderDesk is running. This happens because:

  1. pythonw.exe was being used instead of python.exe for Python operations
  2. pythonw.exe is the "windowless" Python executable intended for GUI applications that don't need a console, where as python.exe is the standard console Python executable that properly handles CLI operations without spawning additional windows
  3. This causes Windows to spawn visible console windows when running pip commands

  ## Solution

  This PR fixes the issue by:

  Changing PYTHON_COMMAND to use python.exe instead of pythonw.exe on Windows

 ## Testing

  - Launch AiderDesk on Windows
  - Keep it running for at least 10-15 minutes to verify no terminal windows pop up periodically
  - Test creating a new project to ensure the setup process doesn't show terminal windows
  - Verify that Aider functionality still works correctly

  This change is Windows-specific (the constant already uses the correct Python binary on other platforms), so it
  shouldn't affect Linux or macOS users.


wladimiiir (approved):

wladimiiir:
Thanks for the contribution.

PR #217, opened 2025-06-16
Fix/progressbar mac

kristian-ziga:


src/main/progress-window.ts
```diff
@@ -0,0 +1,76 @@
...
+      show: false,
+      icon: options.icon,
+      backgroundColor: '#1c2025',
+      webPreferences: {
+        nodeIntegration: true,
+        contextIsolation: false,
+      },
+    });
+
+    this.window.loadFile('src/renderer/src/utils/progress.html');
```
wladimiiir: This should be moved under event handler registration.

src/main/index.ts
```diff
@@ -190,17 +162,18 @@ app.whenReady().then(async () => {
...
   try {
     await performStartUp(updateProgress);
-    updateProgress({
-      step: 'Startup complete',
-      message: 'Everything is ready! Have fun coding!',
-    });
+    progressBar.text = 'Startup complete';
+    progressBar.detail = 'Everything is ready! Have fun coding!';
     progressBar.setCompleted();
     await delay(1000);
```
wladimiiir: I think we can remove this delay as there is enough time for UI to initialize anyways.

src/renderer/src/utils/progress.html
```diff
@@ -0,0 +1,78 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Progress</title>
```
wladimiiir: Change to `Starting AiderDesk...`

wladimiiir (approved):
Great job, thanks :+1:

PR #226, opened 2025-06-17
Feature/clear errors

kristian-ziga:
Added clear-logs command to remove all log messages from chat.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [UNRELEASED]
 
 - revamped starting progress bar
+- added /clear-logs command to clear log messages from chat
 
 ## [0.18.0]
 


src/common/locales/en.json
```diff
@@ -538,6 +538,7 @@
     "architect": "Switch to architect mode. Architect mode is used to plan changes to the code base and execute them",
     "ask": "Switch to ask mode. Ask questions about the code base without editing any files",
     "clear": "Clear the chat history",
+    "clear-logs": "Remove LogMessages from chat",
```
wladimiiir: Change to "Remove only log messages from chat"

src/common/locales/zh.json
```diff
@@ -532,6 +532,7 @@
     "architect": "切换到架构师模式。架构师模式用于规划和执行代码库的更改。",
     "ask": "切换到提问模式。在不编辑任何文件的情况下询问有关代码库的问题。",
     "clear": "清除聊天历史记录",
+    "clear-logs": "",
```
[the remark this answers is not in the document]
wladimiiir: @kristian-ziga 
You can use the suggestion from here.

src/renderer/src/components/project/ProjectView.tsx
```diff
@@ -440,6 +441,11 @@ export const ProjectView = ({ project, modelsInfo, isActive = false }: Props) =>
     }
   };
 
+  const clearLogMessages = () => {
+    setMessages((prevMessages) => prevMessages.filter((message) => !isLogMessage(message)));
+    setProcessing(false);
```
wladimiiir: We don't need to set processing to `false` here.

kristian-ziga: May I ask when it should be used?

wladimiiir: The `processing` state indicates an active AI request or response; therefore, only actions intended to interrupt this state should modify it. I've observed that `clearMessages` currently sets `processing` to `false`, which is likely incorrect and requires verification. 

Regardless, calling `clearLogMessages` while `processing` is `true` would be impossible, as the `PromptField` prevents the command from running.

kristian-ziga: Thanks for explanation. I saw it in clearMessages and that is why I used it.

wladimiiir (approved):

PR #246, opened 2025-07-01
Feat/notification when update ready to install

kristian-ziga:
Added notification when update is ready to be installed


src/renderer/src/pages/Home.tsx
```diff
@@ -46,6 +48,13 @@ export const Home = () => {
   const isDownloading = typeof versions?.aiderDeskDownloadProgress === 'number';
   const showUpdateIcon = isDownloading || isUpdateAvailable || versions?.aiderDeskNewVersionReady;
 
+  useEffect(() => {
+    if (versions?.aiderDeskNewVersionReady && !hasShownUpdateNotification) {
+      showInfoNotification(t('settings.about.updateReadyToBeInstalled'));
```
wladimiiir: Use existing `settings.about.newAiderDeskVersionReady` instead.

PR #260, opened 2025-07-07
Feat/web command to history

kristian-ziga:
/web commands are saved into input history


wladimiiir (changes requested):
Instead of introducing another IPC handler, we should call `project.addToInputHistory` with `/web ${url}`.

kristian-ziga:
Needs the web command with file path pull request to be approved before this.

wladimiiir (approved):

PR #266, opened 2025-07-09
fix: quit when all window closed

kristian-ziga:
Added quick fix. Tried to do it the right way, but could not figure it out.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -226,9 +226,7 @@ app.whenReady().then(async () => {
 });
 
 app.on('window-all-closed', () => {
-  if (process.platform !== 'darwin') {
-    app.quit();
-  }
+  app.quit();
 });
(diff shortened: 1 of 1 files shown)


wladimiiir:
@kristian-ziga 

Can you try to create `closeIpcHandlers` function inside `ipc-handlers.ts` and call it from `beforeQuit`?
The function should call either `ipcMain.removeAllListeners`, if that's enough or individual `ipcMain.removeHandler` for each channel registered by `ipcMain.handle`.

Thanks.

kristian-ziga:
Tried, but still getting EAADRINUSE error and this error in console. UnhandledPromiseRejectionWarning: Error: Attempted to register a second handler for 'load-settings'. I can commit it and push it, but I believe that I did it correctly.

Issue #281, opened 2025-07-21
Feature: clear-done button in todo tool


Slach:
**Is your feature request related to a problem? Please describe.**
As a user i need "delete all done tasks" button in "todo tool"
<img width="459" height="207" alt="Image" src="https://github.com/user-attachments/assets/77c7da32-e22d-48e0-add0-0fa4170e489e" />

wladimiiir:
Thanks for the suggestion, that sounds like a good idea.

wladimiiir:
Implemented in [v0.24.0](https://github.com/hotovo/aider-desk/releases/tag/v0.24.0).

PR #284, opened 2025-07-22
Feat/clear done todo tool

kristian-ziga:
Added button that clears done tasks in todo list.


wladimiiir:
@kristian-ziga 
Sorry, I didn't realize the description of #281. We should actually make the action, so you can clear all todo items regardless if they are done or not. Thanks.

PR #295, opened 2025-07-30
Feat/requesty open router model selectors

kristian-ziga:
Added dropdown list populated with models loaded directly from the Requesty and OpenRouter APIs


src/renderer/src/hooks/useFetchModels.ts
```diff
@@ -0,0 +1,48 @@
...
+interface OpenRouterModel {
+  id: string;
+  // others not needed
+}
+
+interface OpenRouterResponse {
+  data: OpenRouterModel[];
+}
+
+export const useFetchModels = (apiKey: string, url: string) => {
```
wladimiiir: I would rather go for 2 separate hooks here that are specific for both the providers - `useRequestyModels` and `useOpenRouterModels` and have the URL as the internal implementation detail. This way they can be easily reused in other parts if needed.

It's OK that they will have most of the code same, but having it separated makes it also more understandable for other developers.

src/renderer/src/hooks/useFetchModels.ts
```diff
@@ -0,0 +1,48 @@
...
+      } catch (err) {
+        console.error('Error fetching models from ' + url + ':', err);
+        setModels([]);
+      }
+    };
+
+    fetchModels();
+  }, [apiKey]);
+
+  return { toSelectModels: models };
```
wladimiiir: Here we only need to `return models;`, no need for wrapping.

src/renderer/src/hooks/useFetchModels.ts
```diff
@@ -0,0 +1,48 @@
...
+        });
+
+        if (!response.ok) {
+          throw new Error(`HTTP error! status: ${response.status}`);
+        }
+
+        const data: OpenRouterResponse = await response.json();
+
+        const modelIds = data.data?.map((model: { id: string }) => model.id) || [];
+        setModels(modelIds);
```
wladimiiir: The model list should be sorted.

wladimiiir (approved):
Great job, thank you!

PR #297, opened 2025-08-01
Feat/main model remember edit format

kristian-ziga:
aider-desk now remembers editFormat per model per project and reuses it


src/main/ipc-handlers.ts
```diff
@@ -43,12 +43,18 @@ export const setupIpcHandlers = (
     return store.getSettings();
   });
 
-  ipcMain.on('run-prompt', (_, baseDir: string, prompt: string, mode?: Mode) => {
-    void projectManager.getProject(baseDir).runPrompt(prompt, mode);
+  ipcMain.on('run-prompt', async (_, baseDir: string, prompt: string, mode?: Mode) => {
+    const project = projectManager.getProject(baseDir);
+    await project.runPrompt(prompt, mode);
+    // Ensure UI updates and scrolls to bottom after prompt is fully processed
+    mainWindow.webContents.send('ensure-scroll', baseDir);
```
wladimiiir: This probably should not be here.

src/main/ipc-handlers.ts
```diff
@@ -43,12 +43,18 @@ export const setupIpcHandlers = (
...
   });
 
   ipcMain.on('answer-question', (_, baseDir: string, answer: string) => {
-    projectManager.getProject(baseDir).answerQuestion(answer);
+    const project = projectManager.getProject(baseDir);
+    project.answerQuestion(answer);
+    // Ensure UI updates and scrolls to bottom
+    mainWindow.webContents.send('ensure-scroll', baseDir);
```
wladimiiir: Same here.

src/renderer/src/components/project/ProjectBar.tsx
```diff
@@ -47,6 +62,7 @@ export const ProjectBar = React.forwardRef<ProjectTopBarRef, Props>(
     const [sessions, setSessions] = useState<SessionData[]>([]);
     const [sessionPopupVisible, showSessionPopup, hideSessionPopup] = useBooleanState(false);
     const sessionPopupRef = useRef<HTMLDivElement>(null);
+    const [currentEditFormat, setCurrentEditFormat] = useState<EditFormat>('whole');
```
wladimiiir: We cannot assume the default, as this could be different from what Aider uses. It should be `null` and then set it from the main process when Aider sends the models info.

src/renderer/src/components/project/ProjectBar.tsx
```diff
@@ -156,16 +177,18 @@ export const ProjectBar = React.forwardRef<ProjectTopBarRef, Props>(
...
-          onModelsChange({
-            ...modelsData,
-            editFormat: format,
-          });
+      (format: EditFormat, modelToUpdate?: string) => {
+        const targetModel = modelToUpdate || modelsData?.mainModel;
+        if (!targetModel) {
+          return;
         }
+        console.log('joooj', targetModel);
```
wladimiiir: Remove.

src/renderer/src/components/project/ProjectBar.tsx
```diff
@@ -325,13 +348,13 @@ export const ProjectBar = React.forwardRef<ProjectTopBarRef, Props>(
                 removePreferredModel={handleRemovePreferredModel}
               />
             </div>
-            {modelsData?.editFormat && (
+            {modelEditFormats && (
```
wladimiiir: This should check for `currentEditFormat`.

src/renderer/src/components/project/ProjectBar.tsx
wladimiiir: We actually might not need this at all as the `editFormat` is in the `modelsData`.

kristian-ziga: It is not in modelsData anymore, because it is not needed there. I also had issues with using directly modelsData and corresponding editFormat directly from it. This solved the issue.

PR #332, opened 2025-08-14
feat: fixed the unnecessary reload

kristian-ziga:


src/renderer/src/pages/Onboarding.tsx
```diff
@@ -16,7 +17,8 @@ import { showErrorNotification, showInfoNotification } from '@/utils/notificatio
 export const Onboarding = () => {
   const { t } = useTranslation();
   const navigate = useNavigate();
-  const { settings, saveSettings } = useSettings();
+  const { saveSettings } = useSettings();
+  const [localSettings, setLocalSettings] = useState<SettingsData | null>(null);
```
wladimiiir: This should be:
` const [localSettings, setLocalSettings] = useState<SettingsData | null>(settings);`

and there should also be `useEffect` for the update.

```
 useEffect(() => {
    if (settings) {
      setLocalSettings(settings);
    }
  }, [settings]);
```

wladimiiir (approved):
LGTM 👍

PR #372, opened 2025-09-02
feat: added functionality and updated changelog

kristian-ziga:


src/renderer/src/components/ContextFiles/ContextFiles.tsx
```diff
@@ -352,6 +353,17 @@ export const ContextFiles = ({ baseDir, allFiles, showFileDialog }: Props) => {
                   <span
                     className={`select-none text-2xs overflow-hidden ${item.isFolder ? 'context-dimmed' : 'text-text-primary font-semibold'}`}
                     {...(item.isFolder ? { onClick: context.arrowProps.onClick } : {})}
+                    data-tooltip-id="context-files-tooltip"
+                    data-tooltip-content={(() => {
```
wladimiiir: Let's create a named function for this in the component instead of IIFE.

src/renderer/src/components/ContextFiles/ContextFiles.tsx
```diff
@@ -352,6 +353,17 @@ export const ContextFiles = ({ baseDir, allFiles, showFileDialog }: Props) => {
...
+                      if (!tokensInfo?.files || item.isFolder) {
+                        return '';
+                      }
+                      const fileTokenInfo = tokensInfo.files[item.index as string];
+                      if (!fileTokenInfo) {
+                        return '';
+                      }
+                      return `${t('usageDashboard.charts.tokens')}: ${fileTokenInfo.tokens || 0}, ${t('usageDashboard.charts.cost')}: $${(fileTokenInfo.cost || 0).toFixed(5)}`;
```
wladimiiir: It should be possible to return a React node here. Let's show the information so each info is on its own line.

src/renderer/src/components/ContextFiles/ContextFiles.tsx
```diff
@@ -352,6 +353,17 @@ export const ContextFiles = ({ baseDir, allFiles, showFileDialog }: Props) => {
                   <span
                     className={`select-none text-2xs overflow-hidden ${item.isFolder ? 'context-dimmed' : 'text-text-primary font-semibold'}`}
                     {...(item.isFolder ? { onClick: context.arrowProps.onClick } : {})}
+                    data-tooltip-id="context-files-tooltip"
```
wladimiiir, line 356: Let's also add data

wladimiiir, line 370: Let's also add `data-tooltip-delay-show={800}` so there is a higher delay for this tooltip.

src/renderer/src/components/ContextFiles/ContextFiles.tsx
wladimiiir: Actually, let's use the same notation we have in the Session info: "x tokens, $y.yyyyy"

wladimiiir (approved):
Thank you 👍

PR #374, opened 2025-09-02
Feature/font size selector

kristian-ziga:


src/common/api.ts
```diff
@@ -44,6 +44,7 @@ export interface ApplicationAPI {
   saveSettings: (settings: SettingsData) => Promise<SettingsData>;
   saveTheme: (theme: Theme) => Promise<Theme>;
   saveFont: (font: string) => Promise<Font>;
+  saveFontSize: (fontSize: number) => Promise<number>;
```
wladimiiir: Let's not add new API function here and also remove `saveTheme` and `saveFont`. We should also remove all the handlers for the functions as they don't really have any special purpose other then updating settings data. In `SettingsProvider`  we can use `saveSettings` directly to update the values.

wladimiiir: Please also remove them from the `BrowserApi` and `SettingsApi`. Thanks.

wladimiiir (approved):
Thanks 👍

PR #412, opened 2025-10-02
fix: improve API key handling and error messages in agent mode

ohnum:
## Motivation
I hade some issues when using `.aider.conf.yaml` in agent mode. When I switched to agent mode and tried to send a prompt I just received an error:

`Anthropic API key is required in Providers settings or Aider environment variables (ANTHROPIC_API_KEY). Configure credentials in the Settings -> Providers.`

## Claude 4.5 Summary of Changes

1. **Added support for api-key property in `.aider.conf.yml`** (`environment.ts`)
   - Added `readApiKeyFromConfFile()` function that parses the api-key property
   - Handles both single string format: `api-key: gemini=<key>`
   - Handles array format: `api-key: [gemini=<key>, openai=<key>]`
   - Maps provider names correctly (e.g., GOOGLE_API_KEY → gemini)
   - Integrated into the precedence order at steps 5 and 9

2. **Updated agent.ts to use `getEffectiveEnvironmentVariable`**
   - Replaced manual `.env` file loading with `getEffectiveEnvironmentVariable`
   - Now properly loads API keys from all 11 sources including `.aider.conf.yml` files
   - Added proper TypeScript typing for the environment object
   - Removed the cached aiderEnv approach in favor of direct calls

3. **Improved API key validation**
   - Validates API keys before attempting to create the LLM
   - Checks both provider-specific keys and environment variables
   - Provides clear error messages indicating where to configure credentials

### DIFF
--- a/src/main/agent/agent.ts
+++ b/src/main/agent/agent.ts
@@ -1,8 +1,6 @@
 import fs from 'fs/promises';
 import path from 'path';
-import { homedir } from 'os';
 
-import dotenv from 'dotenv';
 import { v4 as uuidv4 } from 'uuid';
 import {
   AgentProfile,
@@ -13,7 +11,6 @@ import {
   McpToolInputSchema,
   PromptContext,
   ProviderProfile,
-  SettingsData,
   ToolApprovalState,
   UsageReportData,
 } from '@common/types';
@@ -56,7 +53,6 @@ import { AIDER_DESK_PROJECT_RULES_DIR } from '@/constants';
 import { Task } from '@/task';
 import { Store } from '@/store';
 import logger from '@/logger';
-import { parseAiderEnv } from '@/utils';
 import { optimizeMessages } from '@/agent/optimizer';
 import { ModelManager } from '@/models/model-manager';
 import { TelemetryManager } from '@/telemetry/telemetry-manager';
@@ -67,7 +63,6 @@ const MAX_RETRIES = 3;
 
 export class Agent {
   private abortControllers: Record<string, AbortController> = {};
-  private aiderEnv: Record<string, string> | null = null;
   private lastToolCallTime: number = 0;
 
   constructor(
@@ -76,19 +71,6 @@ export class Agent {
     private readonly modelManager: ModelManager,
     private readonly telemetryManager: TelemetryManager,
   ) {}
-
-  private invalidateAiderEnv() {
-    this.aiderEnv = null;
-  }
-
-  settingsChanged(oldSettings: SettingsData, newSettings: SettingsData) {
-    const aiderEnvChanged = oldSettings.aider?.environmentVariables !== newSettings.aider?.environmentVariables;
-    const aiderOptionsChanged = oldSettings.aider?.options !== newSettings.aider?.options;
-    if (aiderEnvChanged || aiderOptionsChanged) {
-      logger.info('Aider environment or options changed, invalidating cached environment.');
-      this.invalidateAiderEnv();
-    }
-  }
 
   private async getFilesContentForPrompt(files: ContextFile[], task: Task): Promise<{ textFileContents: string[]; imageParts: ImagePart[] }> {
     const textFileContents: string[] = [];
@@ -534,9 +516,8 @@ export class Agent {
     }
     const effectiveAbortSignal = abortSignal || this.abortControllers[task.project.baseDir]?.signal;
 
-    const llmProvider = provider.provider;
-    const cacheControl = this.modelManager.getCacheControl(profile, llmProvider);
-    const providerOptions = this.modelManager.getProviderOptions(profile.provider, profile.model);
+    const cacheControl = this.modelManager.getCacheControl(profile, provider.provider);
+    const providerOptions = this.modelManager.getProviderOptions(provider.provider, profile.model);
 
     const userRequestMessage: ContextUserMessage = {
       id: promptContext?.id || uuidv4(),
@@ -559,7 +540,7 @@ export class Agent {
       task.addLogMessage('error', `Error reinitializing MCP clients: ${error}`, false, promptContext);
     }
 
-    const toolSet = await this.getAvailableTools(task, profile, llmProvider, contextMessages, resultMessages, effectiveAbortSignal, promptContext);
+    const toolSet = await this.getAvailableTools(task, profile, provider.provider, contextMessages, resultMessages, effectiveAbortSignal, promptContext);
 
     logger.info(`Running prompt with ${Object.keys(toolSet).length} tools.`);
     logger.debug('Tools:', {
@@ -569,7 +550,14 @@ export class Agent {
     let currentResponseId: string = uuidv4();
 
     try {
-      const model = this.modelManager.createLlm(provider, profile.model, await this.getLlmEnv(task));
+      logger.debug('Creating LLM model', {
+        providerId: provider.id,
+        providerName: provider.provider.name,
+        modelName: profile.model,
+      });
+
+      const model = this.modelManager.createLlm(provider, profile.model, settings, task.project.baseDir);
+      logger.debug('LLM model created successfully');
 
       if (!systemPrompt) {
         systemPrompt = await getSystemPrompt(task.project.baseDir, profile);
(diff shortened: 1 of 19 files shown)


wladimiiir:
Thanks for your contribution.

While I plan to delve deeper into this next week, my initial impression is that this extends the coupling between the Agent and its providers by exposing specific API key environment variables. Ideally, this responsibility should be delegated to each provider's `LlmProviderStrategy` interface implementation within `createLlm`. Currently, the function receives an `env` argument externally; however, I believe it should directly use `getEffectiveEnvironmentVariable` within the function, as only the provider's implementation should know which environment variables are needed. Hopefully, my explanation makes sense. It would be ideal, if you could have a look and update accordingly, and let me know what if you see any issues. But even without changes from your end, I think we can still merge it, and I will do the changes manually.

Thanks again for fixing the issue, and I also like the idea of showing the detected API keys in the Settings :+1:

ohnum:
> While I plan to delve deeper into this next week, my initial impression is that this extends the coupling between the Agent and its providers by exposing specific API key environment variables.

Yes, that seems about right.

> Ideally, this responsibility should be delegated to each provider's LlmProviderStrategy interface implementation within createLlm. Currently, the function receives an env argument externally; however, I believe it should directly use getEffectiveEnvironmentVariable within the function, as only the provider's implementation should know which environment variables are needed.

I think the last commits address this issue. The loading of the environment variables has been moved from `agent.ts` to each provider instead.

src/main/agent/agent.ts
```diff
@@ -513,7 +495,7 @@ export class Agent {
     const provider = providers.find((p) => p.id === profile.provider);
     if (!provider) {
       logger.error(`Provider ${profile.provider} not found`);
-      project.addLogMessage('error', 'Selected model is not configured. Select another model and try again.', false, promptContext);
+      project.addLogMessage('error', `Provider ${profile.provider} not found. Please configure the provider in Settings -> Providers.`, false, promptContext);
```
wladimiiir: This should be "... in Model Library".

src/main/agent/agent.ts
```diff
@@ -869,11 +857,7 @@ export class Agent {
       }
 
       logger.error('Error running prompt:', error);
-      if (error instanceof Error && (error.message.includes('API key') || error.message.includes('credentials'))) {
-        project.addLogMessage('error', `${error.message}. Configure credentials in the Settings -> Providers.`, false, promptContext);
-      } else {
-        project.addLogMessage('error', `${error instanceof Error ? error.message : String(error)}`, false, promptContext);
-      }
+      project.addLogMessage('error', `${error instanceof Error ? error.message : String(error)}`, false, promptContext);
```
wladimiiir:
```
if (error instanceof Error && (error.message.includes('API key') || error.message.includes('credentials'))) {
        project.addLogMessage('error', `${error.message}. Configure credentials in the Settings -> Providers.`, false, promptContext);
      }
```

The error should still be checked for these strings, I believe. This is to give a more descriptive message to user.

src/main/agent/agent.ts
```diff
@@ -1013,6 +960,11 @@ export class Agent {
     this.abortControllers[baseDir]?.abort();
   }
 
+  settingsChanged(_oldSettings: SettingsData, _newSettings: SettingsData) {
```
wladimiiir: We can remove this function and its usages.

src/main/models/providers/gemini.ts
```diff
@@ -87,26 +87,45 @@ export const getGeminiAiderMapping = (provider: ProviderProfile, modelId: string
...
+    }
+  }
 
   if (!apiKey) {
     throw new Error('Gemini API key is required in Providers settings or Aider environment variables (GEMINI_API_KEY)');
   }
 
-  const baseUrl = provider.customBaseUrl || env['GEMINI_API_BASE_URL'];
-
-  const providerOverrides = model.providerOverrides as Partial<GeminiProvider> | undefined;
```
wladimiiir: The use of `providerOverrides` should not be removed.

src/main/models/providers/openrouter.ts
```diff
@@ -110,24 +110,27 @@ export const getOpenRouterAiderMapping = (provider: ProviderProfile, modelId: st
...
+      apiKey = effectiveVar.value;
+      logger.debug(`Loaded OPENROUTER_API_KEY from ${effectiveVar.source}`);
+    }
+  }
 
   if (!apiKey) {
     throw new Error('OpenRouter API key is required in Providers settings or Aider environment variables (OPENROUTER_API_KEY)');
   }
 
-  const providerOverrides = model.providerOverrides as Partial<OpenRouterProvider> | undefined;
```
wladimiiir, line 122: The use of `providerOverrides` should not be removed.

src/main/models/providers/requesty.ts
```diff
@@ -110,17 +110,21 @@ export const getRequestyAiderMapping = (provider: ProviderProfile, modelId: stri
...
   if (!apiKey) {
-    throw new Error('Requesty API key is required in Providers settings or Aider environment variables (REQUESTY_API_KEY)');
+    const effectiveVar = getEffectiveEnvironmentVariable('REQUESTY_API_KEY', settings, projectDir);
+    if (effectiveVar) {
+      apiKey = effectiveVar.value;
+      logger.debug(`Loaded REQUESTY_API_KEY from ${effectiveVar.source}`);
+    }
   }
 
-  const providerOverrides = model.providerOverrides as Partial<RequestyProvider> | undefined;
```
wladimiiir, line 118: The use of `providerOverrides` should not be removed.

src/main/models/model-manager.ts
```diff
@@ -480,40 +480,12 @@ export class ModelManager {
     return strategy.getAiderMapping(provider, modelId);
   }
 
-  getModel(providerId: string, modelId: string): Model | undefined {
```
wladimiiir, line 491: I believe, by resolving the conflicts, this function (and some other latest changes) have been removed.

src/main/models/types.ts
```diff
@@ -31,8 +31,9 @@ export interface LlmProviderStrategy {
   // === LLM Creation and Usage Functions ===
   /**
    * Creates a LanguageModel instance for the given provider and model
+   * Each provider is responsible for loading its own credentials using getEffectiveEnvironmentVariable
    */
-  createLlm: (profile: ProviderProfile, model: Model, env: Record<string, string | undefined>) => LanguageModel;
```
wladimiiir: `model: Model` should still be used here, so `providerOverrides` can be utilized

src/renderer/src/components/settings/AiderSettings.tsx
```diff
@@ -107,31 +107,34 @@ export const AiderSettings = ({ settings, setSettings, initialShowEnvVars = fals
...
-            />
-            {!showEnvVars && (
-              <div className="absolute inset-[3px] bottom-[9px] bg-bg-primary-light-strong backdrop-blur-sm flex items-center justify-center rounded-sm">
-                <Button variant="text" color="secondary" onClick={() => setShowEnvVars(true)} className="flex items-center" size="sm">
-                  <HiEye className="mr-2" /> {t('settings.common.showSecrets')}
-                </Button>
-              </div>
-            )}
+        <div className="px-4 py-6 pb-3 space-y-4">
+          <div>
```
wladimiiir: There is additional `div` that does not need to be here. I believe, you had a nice feature here before, showing the API keys, resolved by the config files. Did you decide against it? Thanks.

ohnum:
> Some of the latest changes have been removed probably during the conflict resolution. Please, see my comments.

Ouch, I'll go back and see what I missed.

src/main/models/providers/gemini.ts
ohnum: Fixed

src/main/models/providers/requesty.ts
ohnum: Fixed

src/main/models/types.ts
ohnum: Now uses the same signature as main.

Issue #427, opened 2025-10-15
Question: How to use gpt-5-codex in Aider modes


tomas55:
**Describe the bug**
Invoked Aider prompts failing with openai/gpt-5-codex model

**Environment:**
- **OS:** Ubuntu 24.04.3
- **Version:** 0.35 

**To Reproduce**
Setup agent profile Aider with Power Search with few additional settings:
- Use openai/gpt-5-codex model
- Add codex7 and bright-data MCP servers
- No other settings changed in the agent profile

Use agent mode
When agent executes run_prompt and evaluates prompt for code inclusion, it responds in ***Run prompt in Aider*** with an error:
```
litellm.BadRequestError: OpenAIException - {
  "error": {
    "message": "Unsupported parameter: 'temperature' is not supported with this model.",
    "type": "invalid_request_error",
    "param": "temperature",
    "code": null
  }
}
```
  

**Expected behavior**
Aider prompt should be invoked with gpt-5-codex model by agent

**Additional context**
Aider-Desk Options set:
- Enable prompt caching
- Watch files for changes

wladimiiir:
Thanks for the report.

This is an issue in Aider not having the correct configuration for the `gpt-5-codex` model yet. You can work around it by putting these files in your project folder.

`.aider.models.settings.yml`
```
- name: openai/gpt-5-codex
  edit_format: diff
  weak_model_name: openai/gpt-5-nano
  use_repo_map: true
  use_temperature: false
  accepts_settings: ["reasoning_effort"]
```

`.aider.conf.yml`
```
model-settings-file: .aider.model.settings.yml
```

tomas55:
Thx!

PR #432, opened 2025-10-22
feat: allow monkey patching of models missing in aider

ohnum:
I was using claude-sonnet-4-5 and I thought it was getting expensive and started to investigate why it seemed to cost me more than with claude-sonnet-4. I found that aider doesn't support prompt caching yet for sonnet and haiku on 4.5 so I thought that maybe I could fix it.

The changes proposed here allow for monkey patching the model settings list in aider at runtime. I'm not even sure this is a feature that is wanted for this application, maybe it should be targeted at the aider repo directly instead to update model_settings.yaml. But I thought this could be food for thought.

The logging is a bit ugly since it outputs directly to stderr, but I couldn't figure out a good way of using the logger in the typescript project.

### DIFF
--- a/resources/connector/connector.py
+++ b/resources/connector/connector.py
@@ -22,6 +22,8 @@ from concurrent.futures import ThreadPoolExecutor, Future
 import nest_asyncio
 import litellm
 import types
+from monkeypatch import monkey_patch_model_settings
+from logger import log_info, log_error, log_warning, log_debug
 
 class PromptContext:
   def __init__(self, id: str, group=None):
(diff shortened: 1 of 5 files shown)


wladimiiir:
@ohnum 
Thanks for the effort. I believe that this can be achieved by just overriding the model settings by using the Aider's [model configuration](https://aider.chat/docs/config/adv-model-settings.html). See my comment [here](https://github.com/hotovo/aider-desk/issues/427#issuecomment-3407439510) related to `gpt-5-codex`.

ohnum:
> @ohnum Thanks for the effort. I believe that this can be achieved by just overriding the model settings by using the Aider's [model configuration](https://aider.chat/docs/config/adv-model-settings.html). See my comment [here](https://github.com/hotovo/aider-desk/issues/427#issuecomment-3407439510) related to `gpt-5-codex`.

Nice, I'll close this then.

PR #453, opened 2025-11-01
Fix MCP server import paths to resolve build errors

UmarUzdanov:
- Remove .js extensions from SDK imports
- Change zod/v3 to zod
- Add --external flags to build script to bypass parent Yarn PnP interference

Fixes build failures caused by incorrect module resolution and parent directory Yarn PnP manifest blocking dependency access. Build failures affected all platforms (Windows, macOS, Linux).

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "start": "electron-vite preview",
     "dev": "electron-vite dev",
     "dev:no-hmr": "NO_HMR=true electron-vite dev",
-    "build:mcp": "esbuild src/mcp-server/aider-desk-mcp-server.ts --bundle --platform=node --outdir=out/mcp-server",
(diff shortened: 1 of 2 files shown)


wladimiiir:
@UmarUzdanov 

Thanks for the effort. Can you please explain your reasoning behind this? The build actually works fine, otherwise the application would not bundle.

Additionally, your changes are removing the bundling and thus breaking the `node aider-desk-mcp-server.js` when using outside of the development cycle in the production Electron packed app.

```
node:internal/modules/cjs/loader:1404
  throw err;
  ^

Error: Cannot find module '@modelcontextprotocol/sdk/server/mcp.js'
Require stack:
- /tmp/test-mcp-server.js
    at Function._resolveFilename (node:internal/modules/cjs/loader:1401:15)
    at defaultResolveImpl (node:internal/modules/cjs/loader:1057:19)
    at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1062:22)
    at Function._load (node:internal/modules/cjs/loader:1211:37)
```

Thank you.

PR #497, opened 2025-11-10
feat: add keyboard shortcuts

ohnum:
I wanted to be able to navigate the UI more from the keyboard so I added some keyboard shortcuts. Feel free to add suggestions or add a discussion about what scheme might be more appropriate.

### Project-level shortcuts (in Home.tsx)
Alt/Cmd+[1-9]: Switch to specific project tab (Alt on Windows/Linux, Cmd on macOS)
Cmd/Ctrl+W: Close current project tab
Cmd/Ctrl+T: Open new project dialog
Cmd/Ctrl+U: Open usage dashboard
Cmd/Ctrl+M: Open model library
Cmd/Ctrl+S: Open settings

### Task-level shortcuts (in ProjectView.tsx):
Ctrl+[1-9]: Switch to specific task tab (Ctrl only, works on all platforms)
Cmd/Ctrl+Shift+T: Create new task
Cmd/Ctrl+Shift+W: Delete current task
Cmd/Ctrl+L: Focus text input

### Dialog shortcuts:
Cmd/Ctrl+B: Open file browser (when project dialog is open)
ESC: Close modals (Model Library, Usage Dashboard, Settings, Open Project Dialog)

### This scheme ensures:
On macOS: Cmd for projects, Ctrl for tasks
On Windows/Linux: Alt for projects, Ctrl for tasks
No conflicts with system shortcuts


src/renderer/src/components/common/Accordion.tsx
```diff
@@ -81,20 +81,18 @@ export const Accordion = ({
             : 'max-h-0 opacity-0',
         )}
       >
-        <Activity mode={isOpen ? 'visible' : 'hidden'}>
```
wladimiiir, line 84: Activity should still be used here.

src/renderer/src/components/common/BaseDialog.tsx
```diff
@@ -40,8 +40,8 @@ export const BaseDialog = ({ title, onClose, children, contentClass, footer, wid
         }}
       >
         <div
-          style={{ width: `${width}px` }}
-          className="bg-bg-secondary-light-strongest shadow-2xl rounded-xl border border-bg-tertiary-strong max-h-[90vh] flex flex-col"
+          style={{ width: `${width}px`, height: '90vh' }}
+          className="bg-bg-secondary-light-strongest shadow-2xl rounded-xl border border-bg-tertiary-strong flex flex-col"
```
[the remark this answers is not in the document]
wladimiiir: As mentioned here, we need to use the `max-height` otherwise even small dialogs will expand unnecessarily.

src/renderer/src/components/common/ModalOverlayLayout.tsx
```diff
@@ -12,6 +12,18 @@ type Props = {
 
 export const ModalOverlayLayout = ({ title, onClose, children }: Props) => {
   const { t } = useTranslation();
+
+  useEffect(() => {
```
wladimiiir: Consider using library such as https://github.com/JohannesKlauss/react-hotkeys-hook for managing the hotkeys.

src/renderer/src/components/project/ProjectView.tsx
```diff
@@ -247,6 +255,58 @@ export const ProjectView = ({ project, isActive = false }: Props) => {
...
   );
 
+  useEffect(() => {
+    if (!isActive) {
+      return;
+    }
+
+    const handleKeyDown = (e: KeyboardEvent) => {
+      // Handle Ctrl + [1-9] for direct task switching (Ctrl only, not Cmd)
+      if (e.ctrlKey && !e.metaKey && !e.shiftKey && e.key >= '1' && e.key <= '9') {
```
wladimiiir: I think this should rather go into the TaskSidebar.

src/renderer/src/components/common/Accordion.tsx
ohnum: Fixed

src/renderer/src/components/project/ProjectView.tsx
```diff
@@ -43,11 +47,15 @@ export const ProjectView = ({ project, isActive = false }: Props) => {
       const newTask = await api.createNewTask(project.baseDir);
       // Task will be automatically added via the existing listener
       setActiveTaskId(newTask.id);
+      // Focus the prompt field after a short delay to ensure the task view is rendered
+      setTimeout(() => {
+        focusActiveTaskPrompt();
+      }, 100);
```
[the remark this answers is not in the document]
ohnum: Fixed

src/renderer/src/components/common/BaseDialog.tsx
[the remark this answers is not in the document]
ohnum: Fixed

src/renderer/src/components/project/ProjectView.tsx
```diff
@@ -247,6 +255,58 @@ export const ProjectView = ({ project, isActive = false }: Props) => {
...
+
+    const handleKeyDown = (e: KeyboardEvent) => {
+      // Handle Ctrl + [1-9] for direct task switching (Ctrl only, not Cmd)
+      if (e.ctrlKey && !e.metaKey && !e.shiftKey && e.key >= '1' && e.key <= '9') {
+        e.preventDefault();
+        const taskIndex = parseInt(e.key) - 1;
+        const sortedTasks = optimisticTasks
+          .filter((task) => task.createdAt && task.updatedAt)
+          .filter((task) => !task.archived)
+          .sort((a, b) => b.updatedAt!.localeCompare(a.updatedAt!));
```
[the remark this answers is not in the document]
ohnum: Fixed

src/renderer/src/components/project/ProjectView.tsx
ohnum: Fixed

src/renderer/src/components/common/ModalOverlayLayout.tsx
ohnum: This seems like the better option.

ohnum:
> Thanks for your contribution. I left you some comment. I think we should use https://github.com/JohannesKlauss/react-hotkeys-hook for better management as there is a vision to have these shortcuts configurable via Settings dialog.

Np, I moved to react-hotkeys-hook and added a settings tab where the keys can be configured.

src/renderer/src/components/common/ModalOverlayLayout.tsx
```diff
@@ -12,6 +13,9 @@ type Props = {
 
 export const ModalOverlayLayout = ({ title, onClose, children }: Props) => {
   const { t } = useTranslation();
+
+  useHotkeys('escape', () => onClose(), { enableOnFormTags: true, enableOnContentEditable: true });
```
wladimiiir: I don't think this is a good idea adding it into this Layout. This is a component that can contain dialogs in it, so in case you will have dialog open, press Escape, the whole layout gets closed. It could be maybe solved by using different scope when a dialog is open.

src/renderer/src/components/settings/SettingsDialog.tsx
```diff
@@ -58,7 +58,18 @@ export const SettingsDialog = ({ onClose, initialTab = 0, initialAgentProfileId,
       void api.reloadMcpServers(originalSettings.mcpServers || {});
     }
     onClose();
-  };
+  }, [originalSettings, localSettings, i18n, api, setTheme, setFont, setFontSize, onClose]);
+
+  useEffect(() => {
+    const handleKeyDown = (e: KeyboardEvent) => {
+      if (e.key === 'Escape') {
```
wladimiiir: Adding escape to this dialog is not very safe to do, unless there is a mechanism that will prevent it from happening when there are some changes made, otherwise user can lose them easily. Maybe a single check if the `originalSettings` and `localSettings` are unchanged would solve it.

src/renderer/src/pages/Home.tsx
```diff
@@ -105,57 +108,143 @@ export const Home = () => {
...
+      const key = e.key;
+      const index = parseInt(key) - 1;
+      switchToProjectByIndex(index);
+    },
+    { scopes: 'home', enableOnFormTags: true, enableOnContentEditable: true },
+    [openProjects, activeProject, setActiveProject, switchToProjectByIndex],
+  );
+
+  // Ctrl+Tab cycling
+  useHotkeys(
```
wladimiiir, line 258: Can you please retest this functionality? It seems it stopped working.

src/renderer/src/pages/Home.tsx
wladimiiir: This needs to be increased as a new tab has been added before it.

src/renderer/src/components/common/ModalOverlayLayout.tsx
ohnum: I will try to change the behavior so that escape goes back one level from where you currently are.

src/renderer/src/components/settings/SettingsDialog.tsx
ohnum: I'll try to deal with it similar to ModalOverlayLayout and add auto-saving.

src/renderer/src/pages/Home.tsx
ohnum: Retested and works now.

src/renderer/src/components/ModelLibrary/ModelDialog.tsx
```diff
@@ -30,33 +31,142 @@ export const ModelDialog = ({ model, providers, onSave, onCancel }: Props) => {
...
-      });
-      setProviderOverrides(model.providerOverrides || {});
-    } else {
-      setFormData({
-        id: '',
-        providerId: providers[0]?.id || '',
-      });
-      setProviderOverrides({});
-    }
+    const existingModel = initialModelRef.current;
```
wladimiiir: These changes seems unnecessary and unrelated to the keyboard shortcuts. Can we rather not do them in this PR?

wladimiiir:
@ohnum 
I would like to exclude the changes related to `ModelLibrary` from the PR - `ProviderDialog`, changes in `ModelDialog`. Although, it might be a good idea to do some of them, I would like to keep the PR focused on one thing, so it's better for user acceptance testing. Thank you :pray:

ohnum:
> @ohnum I would like to exclude the changes related to `ModelLibrary` from the PR - `ProviderDialog`, changes in `ModelDialog`. Although, it might be a good idea to do some of them, I would like to keep the PR focused on one thing, so it's better for user acceptance testing. Thank you 🙏

It makes sense. I was debating whether or not to include them because I could feel how the scope kept increasing. But I also thought it was a neat solution to the espace issue you mentioned. But I can go back and try to figure out something else.

wladimiiir:
I see. I think for now, we can just not include the close on escape for `ModelOverlayLayout` in this PR and resolve it as a separate enhancement.

wladimiiir:
@ohnum 
Thanks for the latest update. I've added Vitest configuration and unit tests for the files you're about to update. Could you please update your PR with these changes and verify that `npm run test` passes? I'd like to merge the PR this week. Thank you.

ohnum:
I added some tests (`npm run test` passes) and fixed some linting issues.

PR #603, opened 2026-01-09
feat: Add terminal shell selection to settings

jutaz:
Add the ability to select an available shell for the integrated terminal. This includes:

- A new `getAvailableShells` IPC event to retrieve system shells.
- A new `terminal` section in `SettingsData` and default settings.
- Updates to localization files for the new terminal settings.
- A new `TerminalSettings` component for the UI.
- Integration into the main `Settings` page.
- Updates to the `TerminalManager` to use the selected shell.


src/common/locales/en.json
wladimiiir, line 1: Seems like there has been unwanted changes in the file.

src/common/locales/zh.json
wladimiiir, line 1: Also in here, lots of the unnecessary changes.

src/main/utils/shell.ts
```diff
@@ -205,6 +188,72 @@ class ShellDetector {
...
+        if (fs.existsSync(shellPath) && !seen.has(shellPath)) {
+          const name = path.basename(shellPath);
+          shells.push({ path: shellPath, name });
+          seen.add(shellPath);
+        }
+      }
+
+      // Also check /etc/shells
+      try {
+        if (fs.existsSync('/etc/shells')) {
```
wladimiiir, line 238: This does not seem like a good approach. It returns bunch of commands that are not really shells in my system.

Image

src/renderer/src/components/settings/TerminalSettings.tsx
```diff
@@ -0,0 +1,139 @@
...
+            </div>
+
+            {/* Custom Path */}
+            <div className="space-y-2">
+              <RadioButton
+                id="shell-custom"
+                name="shell-mode"
+                value="custom"
+                checked={mode === 'custom'}
+                onChange={() => handleModeChange('custom')}
```
wladimiiir, line 118: Does not seem clicking on custom mode is working.

Image

src/main/utils/shell.ts
```diff
@@ -74,7 +73,7 @@ class ShellDetector {
     const envShell = process.env.SHELL;
     if (envShell && fs.existsSync(envShell)) {
       const name = path.basename(envShell);
-      return { path: envShell, name, args: this.getShellArgs(name) };
```
wladimiiir: I don't think removing this is a good idea. Is this fixing some issue?

wladimiiir (changes requested):
Thanks for your contribution.

I found some issues during the testing. Could you please address them? Thank you.

src/common/locales/zh.json
jutaz: yep, my bad. Will fix.

PR #604, opened 2026-01-09
Feature/litellm provider

jutaz:
Integrates [LiteLLM](https://litellm.ai) as a new LLM provider, allowing users to leverage its capabilities for AI model interactions. This includes adding LiteLLM to common agent types, enabling model discovery, and providing UI components for configuration.

Mostly this just works like an openai provider with the exception of using a different endpoint to fetch the models - this allows us to load the model info properly, get context window sizes, etc.
LiteLLM allows load-balancing, where multiple models can be aliased to same name, in which case the logic is to take the smaller context window and highest cost - i.e. the "worst-case scenario".


jutaz:
Yep! Sorry about that.

PR #615, opened 2026-01-15
Bugfix - Do not close/terminate terminal when switching tasks.

jutaz:
This may have been intentional, but submitting this in case it wasn't. Allows to have longer-running tasks, which is useful when testing what agent has done while replying to another task in the same or different project.


src/renderer/src/components/project/ProjectView.tsx
```diff
@@ -415,9 +417,18 @@ export const ProjectView = ({ project, isActive = false, showSettingsPage }: Pro
...
+    const isLoadingOrLoaded = taskStateMap.get(task.id)?.loading || taskStateMap.get(task.id)?.loaded;
+    const isRecentlyActive = !lastActiveAt || lastActiveAt.getTime() + 1000 * 60 * 15 > Date.now();
+
+    return isActive || (isLoadingOrLoaded && isRecentlyActive);
+  });
+
   return (
     <TaskProvider baseDir={project.baseDir} tasks={tasks}>
-      <div className="h-full w-full bg-gradient-to-b from-bg-primary to-bg-primary-light relative">
+      <div className="h-full w-full bg-gradient-to-b from-bg-primary to-bg-primary-light relative" style={{ contentVisibility: isActive ? 'visible' : 'hidden', zIndex: isActive ? 1 : 0 }}>
```
wladimiiir: I don't think this change is needed. This is already done one level higher in Home component.

wladimiiir:
@jutaz 
When switching between tasks, does the Terminal keep its internal state (previously ran output, etc.)?

wladimiiir:
This should be fixed by the latest changes.

PR #622, opened 2026-01-19
feat: add Bun runtime support with cross-platform binary builds

pcfreak30:
Add comprehensive support for building and running AiderDesk as compiled Bun binaries across multiple platforms (Linux x64/arm64, macOS x64/arm64, Windows x64).

Key changes:
- Add bunup.config.ts for multi-platform Bun binary compilation
- Implement embedded resource system with mutex-protected I/O operations
- Add bun-resources.ts for resource extraction, caching, and renderer file serving
- Refactor data-manager.ts with adapter pattern supporting bun:sqlite and better-sqlite3
- Update startup and server initialization for Bun binary environment
- Add platform-specific binary and resource path handling
- Add build scripts for Bun: build:bun, build:bun:all, and platform-specific targets
- Add dependencies: bunup, @types/bun, simple-git-hooks
- Update TypeScript configurations for Bun types and path mappings

Notes:
* `simple-git-hooks` is forced in b/c `bunup` makes some weird devtools assumptions and errors if its not in `$PATH`. Added this simply to ensure no build issues.
* A CWD workaround is required with bun and due to that all bun logic is isolated to one file and uses the mutex system to prevent race conditions with changing `CWD`. Probably not bulletproof, but it should defensively guard against most things.


src/main/data-manager/data-manager.ts
```diff
@@ -61,12 +177,20 @@ export class DataManager {
    * @param usageReport The usage report data.
    * @param content The content of the message.
    */
-  public saveMessage(id: string, type: 'tool' | 'assistant', project: string, model: string, usageReport: UsageReportData | undefined, content: unknown): void {
+  public async saveMessage(
```
wladimiiir, line 174: The function has been made async, but the callers were not updated. Consider using `await` or `void` there.

src/main/data-manager/data-manager.ts
```diff
@@ -106,14 +230,23 @@ export class DataManager {
...
   }
 
-  public queryUsageData(from: Date, to: Date): UsageDataRow[] {
+  /**
+   * Query usage data between two dates.
+   * @param from Start date
+   * @param to End date
+   * @returns Array of usage data rows
+   */
+  public async queryUsageData(from: Date, to: Date): Promise<UsageDataRow[]> {
```
wladimiiir, line 240: Please verify that UsageDashboard works after the change.

src/main/bun-resources.ts
```diff
@@ -0,0 +1,362 @@
...
+    await access(uvPath);
+    await access(connectorPath);
+    logger.debug('[Bun I/O] Resources already extracted');
+    return;
+  } catch {
+    // One or more files don't exist, proceed with extraction
+  }
+
+  // List of text resources to extract (must match what's embedded in build script)
+  const textResources = [
```
wladimiiir: This is not very maintainable approach. We will need to manage this manually whenever anything changes in the resources.

src/main/start-up.ts
```diff
@@ -58,9 +60,21 @@ const setupAiderConnector = async (cleanInstall: boolean, updateProgress?: Updat
   }
 
   // Copy connector.py from resources
-  const sourceConnectorPath = path.join(RESOURCES_DIR, 'connector/connector.py');
   const destConnectorPath = path.join(AIDER_DESK_CONNECTOR_DIR, 'connector.py');
-  fs.copyFileSync(sourceConnectorPath, destConnectorPath);
+
+  if (isBunBinary()) {
```
wladimiiir: There is a logic in `paths.ts` file to determine value `RESOURCE_DIR`. I think the `isBunBinary` conditioning should go there instead.

src/main/start-up.ts
```diff
@@ -191,20 +205,37 @@ const setupMcpServer = async () => {
   // Copy all files from the MCP server directory
   const sourceMcpServerDir = path.join(RESOURCES_DIR, 'mcp-server');
 
-  if (fs.existsSync(sourceMcpServerDir)) {
-    const files = fs.readdirSync(sourceMcpServerDir);
+  if (isBunBinary()) {
```
wladimiiir, line 190: It would be better to extract this as separate file with all the special start up logic for Bun.

wladimiiir: Or maybe have helper function `copyResourceFile` that will decide how to copy the file, so the this file does not have to know about the Bun.

package.json
```diff
@@ -20,11 +20,19 @@
...
     "dev": "electron-vite dev",
     "dev:no-hmr": "NO_HMR=true electron-vite dev",
-    "build:mcp": "esbuild src/mcp-server/aider-desk-mcp-server.ts --bundle --platform=node --outdir=out/mcp-server",
-    "prebuild:server": "electron-vite build && npm run build:mcp",
+    "prebuild:server": "npm run build:renderer && npm run build:mcp",
     "build:server": "vite build --config vite.config.server.ts && mkdir -p out/resources && cp -r resources/* out/resources",
     "postbuild:server": "npm rebuild",
-    "build": "npm run typecheck && electron-vite build && npm run build:mcp",
+    "build": "npm run typecheck && npm run build:renderer && npm run build:mcp",
+    "build:renderer": "electron-vite build",
```
wladimiiir: This is not really only renderer but the whole app with renderer and main.

tsconfig.json
```diff
@@ -1,5 +1,16 @@
 {
   "files": [],
+  "compilerOptions": {
+    "baseUrl": ".",
+    "paths": {
+      "@/*": [
+        "src/main/*"
```
wladimiiir: This is not correct in the context of `tsconfig.web.json`.

tsconfig.node.json
```diff
@@ -8,7 +8,7 @@
   ],
   "compilerOptions": {
     "composite": true,
-    "types": ["electron-vite/node"],
+    "types": ["electron-vite/node", "bun"],
```
wladimiiir, line 11: Can we have separate `tsconfig.bun.json` file instead to keep this clean?

tsconfig.json
pcfreak30: @wladimiiir this is kind of required. `bunup` removed needing a custom build script but I cannot reasonably get it to use a custom tsconfig that doesnt create weird build time race conditions.

I have tried and it was better to give in and put the alias in the default config.

tsconfig.node.json
pcfreak30: @wladimiiir no, bunup requires the default config and I could not get it to listen otherwise... despite me RE'ing the code and figuring out where it passes the tsconfig data.

pcfreak30: > @wladimiiir no, bunup requires the default config and I could not get it to listen otherwise... despite me RE'ing the code and figuring out where it passes the tsconfig data.

This problem seems to be resolved and im reviewing the other issues.

pcfreak30:
> In `bunup.config.ts`, the `PLATFORM_INCLUDES` map includes `'linux'` for `linux-x64` but not for `linux-arm64` — if the generic `linux` bin directory contains shared ARM-compatible binaries, this asymmetry would silently exclude them from arm64 builds. Worth verifying this is intentional and not an oversight.
> 
> The `readDirRecursive` function uses `fullPath.replace(baseDir + '/', '')` to compute relative paths, which is fragile on Windows where the path separator is `\` rather than `/`. Since `windows-x64` is a supported target platform (and Bun itself runs on Windows), this could produce incorrect keys in the `files` map. Using `path.relative(baseDir, fullPath)` would be more robust.
> 
> The function signature `readDirRecursive(dir, baseDir = dir, addResourcesPrefix = false, platform = 'linux-x64')` is missing TypeScript type annotations despite being in a `.ts` file — inconsistent with the rest of the file and loses type safety for the `platform` parameter, which could silently accept an invalid string that then falls through to the `DEFAULT_PLATFORM` fallback in `shouldIncludeResource` without any warning.

This PR is likely dead b/c the same can be achieved now in node directly and bun has an upstream bug pending a fix in a sea of 3k claude PR's.

So I am not going to wait on bun.

PR #625, opened 2026-01-20
Add ability to disable thinking with glm-4.7 for zai-plan provider - This isn't a complete idea, I need to think it through.

aguyinatx:
- Add includeThoughts and thinking properties to ZaiPlanProvider interface
- Implement getProviderOptions function for zai-plan provider
- Add thinking toggle control to ZaiPlanParameters UI component
- Add i18n translations for thinking control
- Add providerOverrides property to AgentProfile interface
- Fix type checking errors for model manager

### DIFF
--- a/src/common/agent.ts
+++ b/src/common/agent.ts
@@ -251,8 +251,7 @@ export const isOpenCodeProvider = (provider: LlmProviderBase): provider is OpenC
 export interface ZaiPlanProvider extends LlmProviderBase {
   name: 'zai-plan';
   apiKey: string;
-  includeThoughts?: boolean;
-  thinking?: boolean;
+  thinkingEnabled?: boolean;
 }
 export const isZaiPlanProvider = (provider: LlmProviderBase): provider is ZaiPlanProvider => provider.name === 'zai-plan';
 
@@ -693,7 +692,6 @@ export const getDefaultProviderParams = <T extends LlmProvider>(providerName: Ll
       provider = {
         name: 'zai-plan',
         apiKey: '',
-        includeThoughts: true,
       } satisfies ZaiPlanProvider;
       break;
     case 'minimax':
--- a/src/common/types.ts
+++ b/src/common/types.ts
@@ -417,11 +417,9 @@ export interface AgentProfile {
   useMemoryTools: boolean;
   useSkillsTools: boolean;
   customInstructions: string;
-  thinking?: boolean;
   subagent: SubagentConfig;
   isSubagent?: boolean; // flag to indicate if this profile is being used as a subagent
   ruleFiles?: string[]; // Array of absolute paths to rule files for this agent profile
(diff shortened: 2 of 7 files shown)


package-lock.json
wladimiiir, line 1: Please, revert any changes in the `package-lock.json` as there has been nothing changed in `package.json`.

src/common/agent.ts
```diff
@@ -251,6 +251,8 @@ export const isOpenCodeProvider = (provider: LlmProviderBase): provider is OpenC
 export interface ZaiPlanProvider extends LlmProviderBase {
   name: 'zai-plan';
   apiKey: string;
+  includeThoughts?: boolean;
+  thinking?: boolean;
```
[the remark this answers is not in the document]
wladimiiir: We really don't need both. Let's have `thinkingEnabled?: boolean` property with the default value `true`. This means that all the logic must check for `thinkingEnabled === false` for it to be disabled. Undefined value means `thinking` is enabled in the `providerOptions` as well in the UI.

src/common/types.ts
```diff
@@ -417,9 +417,11 @@ export interface AgentProfile {
   useMemoryTools: boolean;
   useSkillsTools: boolean;
   customInstructions: string;
+  thinking?: boolean;
```
wladimiiir: This should not be agent profile specific. We need to keep it on provider/model level. Please remove.

src/common/types.ts
```diff
@@ -417,9 +417,11 @@ export interface AgentProfile {
...
   subagent: SubagentConfig;
   isSubagent?: boolean; // flag to indicate if this profile is being used as a subagent
   ruleFiles?: string[]; // Array of absolute paths to rule files for this agent profile
+  providerOverrides?: Record<string, unknown>;
```
wladimiiir: Let's not add this, as there is no use case for this.

src/main/models/providers/zai-plan.ts
```diff
@@ -114,15 +108,32 @@ const getZaiPlanModelInfo = (_provider: ProviderProfile, modelId: string, allMod
   return allModelInfos[fullModelId];
 };
 
-// === Complete Strategy Implementation ===
+const getZaiPlanProviderOptions = (llmProvider: LlmProvider, model: Model, profile?: AgentProfile): SharedV2ProviderOptions | undefined => {
+  if (isZaiPlanProvider(llmProvider)) {
+    const modelProviderOverrides = model.providerOverrides as Partial<ZaiPlanProvider> | undefined;
+
+    const includeThoughts = profile?.thinking ?? modelProviderOverrides?.includeThoughts ?? llmProvider.includeThoughts ?? true;
```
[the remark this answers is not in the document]
wladimiiir: Let's only have the `thinkingEnabled` property, the agent profile is not needed here and should be removed from the whole chain of calls.

PR #632, opened 2026-01-25
Chore - Optimize tooltips

jutaz:
Optimize tooltips. Tooltips were using dom observations, which did slow down UI quite significantly, especially with many messages being generated. This addresses that by creating a registry for tooltips. We now use selectors for an anchor, which allows us to not have observers running.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,15 @@
 ## [UNRELEASED]
 
 - hiding the subtasks when parent task is archived
+- skipping model loading for disabled providers
+- updated visuals of model selectors
+- Claude Agent SDK provider only available when claude executable is present
+- added action to handoff a todo item to a new task
+
+## [0.50.1]
+
+- setting the executably path manually for Claude Agent SDK provider in Electron
+- fixed internal TypeError when processing the response
 
 ## [0.50.0]
 
--- a/src/common/locales/en.json
+++ b/src/common/locales/en.json
@@ -926,7 +926,8 @@
     "cancelTodo": "Cancel",
     "noTodos": "No todo items",
     "confirmDelete": "Are you sure you want to delete this todo item?",
-    "clearAllTodos": "Clear all todo items"
+    "clearAllTodos": "Clear all todo items",
+    "handoffTodo": "Handoff to new task"
   },
   "tool": {
     "noDescription": "No description",
--- a/src/common/locales/zh.json
+++ b/src/common/locales/zh.json
@@ -902,7 +902,8 @@
     "cancelTodo": "取消",
     "noTodos": "没有待办事项",
     "confirmDelete": "确定要删除这个待办事项吗？",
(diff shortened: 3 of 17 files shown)


src/renderer/src/components/message/UsageInfo.tsx
```diff
@@ -2,31 +2,34 @@ import { useTranslation } from 'react-i18next';
...
 export const UsageInfo = ({ usageReport, className }: Props) => {
+  const { id, updateTooltipContent } = useTooltip({ content: '' }, 'usage-info-tooltip');
   const { t } = useTranslation();
 
   return (
     <div className={className}>
       <div className="flex items-center gap-3 px-2 text-2xs text-text-muted group-hover:text-text-secondary transition-colors">
         {usageReport.cacheWriteTokens && usageReport.cacheWriteTokens > 0 ? (
-          <span className="flex items-center gap-1" data-tooltip-id="usage-info-tooltip" data-tooltip-content={t('responseMessage.cacheWriteTokens')}>
+          <span className="flex items-center gap-1" data-tooltip-id={id} onMouseEnter={() => updateTooltipContent(t('responseMessage.cacheWriteTokens'))}>
```
wladimiiir: I am not sure about this. From simple `data-tooltip-content` to the custom `onMouseEnter` function that is created on each render. There must a better and easily resuable pattern.

src/renderer/src/components/common/StyledTooltip.tsx
```diff
@@ -5,14 +5,17 @@ type Props = {
...
 
-export const StyledTooltip = ({ id, content, maxWidth = '300px' }: Props) => (
+export const StyledTooltip = ({ id, content, maxWidth = '300px', disableInternalActivator = false }: Props) => (
   <Tooltip
     id={id}
     className="!bg-bg-primary-light !text-text-secondary !text-2xs !py-1 !px-2 !opacity-100 !rounded-md z-50 whitespace-pre-wrap select-none"
     border="1px solid #495057"
     delayShow={200}
+    // @ts-expect-error react-tooltip types missing disableInternalActivator
+    disableInternalActivator={disableInternalActivator}
```
wladimiiir: There does not seem any mention about `disableInternalActivator` in the source codes of `react-tooltip` library. Can you give me source of this info and what exactly is it supposed to do? Thanks.

wladimiiir:
@jutaz 
Thanks for the effort. Since you've identified `react-tooltip` as the performance bottleneck, rather than "patching" it with a custom solution, why not replace it with another library such as [@radix-ui/react-tooltip](https://www.radix-ui.com/primitives/docs/components/tooltip) or [@floating-ui/react](https://floating-ui.com/docs/useFloating)? Do you think you could look into implementing one of those solutions instead? Let me know your thoughts, thank you.

jutaz:
Not a bad idea - let me look 👍

PR #636, opened 2026-01-28
Bugfix - Fix terminal opening in main dir for subtasks

jutaz:
Ensures that terminal, when opened from a subtask, will open in a worktree if parent is also in a subtask. This just basically makes it more consistent as subtasks are part of the worktrees anyways.


wladimiiir:
Looking at this, I have just realized that I fixed the same issue after I have discovered it. Thanks for the effort, we can close this PR.

PR #662, opened 2026-02-10
Support structured connector message content

### LINKED ISSUE #661: Bug: Images are not passed to Aider in "Code" mode
### Bug Report: Images are not passed to Aider in "Code" mode

**Summary:**
In **Code** mode (Aider integration), images added to the task context are not transmitted to the LLM. While multimodal content works correctly in **Agent** mode, the visual context is lost during the data transfer from the TypeScript main process to the Python connector when using Aider.

**Steps to Reproduce:**
1. Open any project in AiderDesk.
2. Switch to **Code** mode.
3. Add an image file (e.g., `screenshot.png`) to the context via the sidebar or by pasting it from the clipboard.
4. Enter a prompt related to the image (e.g., *"Implement the layout based on this image"*).
5. Observe the model's response.

**Expected Result:**
Aider receives the image as part of the multimodal context (via `abs_image_fnames` or a multimodal message structure), and the model responds with awareness of the visual data.

**Actual Result:**
The model either states it cannot see any images or fails to process the file because it attempts to read it as a text/binary blob instead of a visual input.

---

### **Technical Analysis:**

**1. Content Filtering in TypeScript:**
The `toConnectorMessages` method in `src/main/task/context-manager.ts` prepares history for the Python connector. It relies on the `extractTextContent` utility, which explicitly strips out everything that isn't plain text:
```typescript
// src/common/utils.ts
export const extractTextContent = (content: unknown): string => {
  // ...
  if (Array.isArray(content)) {
    return content.filter(isTextContent) // <--- This removes image objects (image_url/base64)
      .map((c) => (typeof c === 'string' ? c : c.text))
      .join('\n\n');
  }
}
```

**2. Missing File Classification in the Connector:**
In `resources/connector/connector.py`, the `clone_coder` function only categorizes files into two sets: `abs_fnames` (editable) and `abs_read_only_fnames` (read-only). 
Aider specifically requires images to be placed in a separate `abs_image_fnames` set to trigger multimodal processing. Currently, this set is not even initialized or populated in the connector.

**3. Simplified Message Protocol:**
The connector currently assumes that the `content` of a message is always a simple string:
```python
# resources/connector/connector.py
coder.done_messages = [dict(role=msg['role'], content=msg['content']) for msg in messages]
```
If a multimodal array (text + image) is passed, it might cause schema validation errors in LiteLLM or be ignored by Aider's internal logic.

---

### **Proposed Fix:**
1.  **TypeScript:** Update `toConnectorMessages` to preserve image parts in the content array when sending data to the connector.
2.  **Python:** Modify `clone_coder` in `connector.py` to use `is_image_file(file_path)` and populate `coder.abs_image_fnames`.
3.  **Python:** Ensure the connector can handle multimodal content objects in the `messages` list instead of assuming strings.

### Steps to reproduce

Open AiderDesk and load any project.
Ensure you are in "Code" mode (using the mode selector below the prompt field).
Paste an image from your clipboard directly into the message editor field (Ctrl+V or Cmd+V).
Observe the UI:
A new image file (e.g., image-001.png) appears in the "Context Files" list in the right sidebar.
The file is physically created in the .aider-desk/tmp/images/ directory.
Ask the model about the file: Type a prompt like "Is there a file named image-001.png in our chat context? Can you see it?".
Observe the response: Read the model's reply.

### Expected behavior

_No response_

### Operating System

Windows

### Version

v0.52.0

### Screenshots

_No response_

### Additional context

_No response_

mykolarudenko:
## Summary
- Add ConnectorMessage/ConnectorMessageContent types so connector messages can carry structured content.
- Normalize context message content to preserve images and convert code blocks to fenced text when needed.
- Handle empty message/file inputs safely for token info and connector messaging.

Fixes #661

## Testing
- npm run test -- --no-color

### DIFF
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -11,6 +11,12 @@ type TextContent =
       text: string;
     };
 
+type CodeBlockContent = {
+  type: 'code-block';
+  code?: unknown;
+  language?: unknown;
+};
+
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const isTextContent = (content: any): content is TextContent => content?.type === 'text' || typeof content === 'string';
 
@@ -20,10 +26,25 @@ export const extractTextContent = (content: unknown): string => {
   }
 
   if (Array.isArray(content)) {
-    return content
-      .filter(isTextContent)
-      .map((c) => (typeof c === 'string' ? c : c.text))
-      .join('\n\n');
+    const textParts: string[] = [];
+
+    for (const part of content) {
+      if (isTextContent(part)) {
+        textParts.push(typeof part === 'string' ? part : part.text);
+        continue;
+      }
+
+      if (typeof part === 'object' && part !== null && 'type' in part && (part as { type?: string }).type === 'code-block') {
+        const codePart = part as CodeBlockContent;
+        const code = typeof codePart.code === 'string' ? codePart.code : '';
+        const language = typeof codePart.language === 'string' ? codePart.language : '';
+        if (code.trim().length > 0 || language.trim().length > 0) {
+          textParts.push(formatCodeBlockText(code, language));
+        }
+      }
+    }
+
+    return textParts.join('\n\n');
   }
 
   if (typeof content === 'object' && content !== null && 'content' in content) {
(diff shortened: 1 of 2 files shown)


wladimiiir:
@mykolarudenko 

Thanks for the contribution, but I don't believe this fixes the issue. The `abs_image_fnames` property you are referencing isn't actually part of Aider's `Coder` class; let me know if I'm missing something there.

Additionally, some of the changes made to the message processing might not be necessary for this fix. I am also unsure about the `code-block` part, as the `AssistantMessage` type doesn't seem to have such content part type. Could you clarify where that is coming from?

mykolarudenko:
> Thanks for the contribution, but I don't believe this fixes the issue. The `abs_image_fnames` property you are referencing isn't actually part of Aider's `Coder` class; let me know if I'm missing something there.
> 

Converted to a draft. Yes, sorry. I fixed the issue on my local install. Just a few lines of code. Models started reporting images.  Then I tried to share. But in the process my bots started getting suggestions from the github bots and now I see a complete mess there. Need to undo the changes and review/test again.

PR #673, opened 2026-02-18
Feature - gemini cli provider

jutaz:
Similarly to how `claude` is implemented, this PR adds support for Gemini CLI. Allows setting the google cloud project ID via UI, since that's something that's common with gemini cli due to the ability to authenticate with multiple projects. Tested this out, seems to be performing quite well across the board.

### DIFF
--- a/patches/ai-sdk-provider-gemini-cli+1.5.1.patch
+++ b/patches/ai-sdk-provider-gemini-cli+1.5.1.patch
@@ -1,0 +1,373 @@
+diff --git a/node_modules/ai-sdk-provider-gemini-cli/dist/index.mjs b/node_modules/ai-sdk-provider-gemini-cli/dist/index.mjs
+index de27b24..b04605e 100644
+--- a/node_modules/ai-sdk-provider-gemini-cli/dist/index.mjs
++++ b/node_modules/ai-sdk-provider-gemini-cli/dist/index.mjs
+@@ -1,8 +1,10 @@
+ var __require = /* @__PURE__ */ ((x) => typeof require !== "undefined" ? require : typeof Proxy !== "undefined" ? new Proxy(x, {
+   get: (a, b) => (typeof require !== "undefined" ? require : a)[b]
+ }) : x)(function(x) {
+-  if (typeof require !== "undefined") return require.apply(this, arguments);
+-  throw Error('Dynamic require of "' + x + '" is not supported');
++  if (typeof require !== "undefined") {
++    return require.apply(this, arguments);
++  }
++  throw Error("Dynamic require of \"" + x + "\" is not supported");
+ });
+ 
+ // src/gemini-provider.ts
+@@ -18,6 +20,7 @@ import {
+   createContentGeneratorConfig,
+   AuthType
+ } from "@google/gemini-cli-core";
++
+ async function initializeGeminiClient(options, modelId) {
+   let authType;
+   if (options.authType === "api-key" || options.authType === "gemini-api-key") {
+@@ -173,6 +176,7 @@ function mapPromptToGeminiFormat(options) {
+   }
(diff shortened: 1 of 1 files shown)


src/main/models/providers/gemini-cli.ts
```diff
@@ -0,0 +1,122 @@
...
+    model: model.id,
+    projectId,
+  });
+
+  // Set GOOGLE_CLOUD_PROJECT env var if projectId is configured (for organization/enterprise accounts)
+  if (projectId) {
+    process.env.GOOGLE_CLOUD_PROJECT = projectId;
+  }
+
+  const gemini = createGeminiProvider({ authType: 'oauth-personal' });
```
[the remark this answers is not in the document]
jutaz: There isn't a way, not at least that I can see for this.

src/common/agent.ts
```diff
@@ -326,6 +335,7 @@ export const DEFAULT_PROVIDER_MODELS: Partial<Record<LlmProviderName, string>> =
   'claude-agent-sdk': 'sonnet',
   deepseek: 'deepseek-chat',
   gemini: 'gemini-3-pro',
+  'gemini-cli': 'gemini-2.5-flash',
```
wladimiiir, line 338: Why not `gemini-3-pro-preview`?

src/main/agent/agent.ts
```diff
@@ -1046,9 +1046,26 @@ export class Agent {
           }
         }
 
+        // Helper to detect non-retryable provider errors
+        const isNonRetryableProviderError = (error: unknown, _providerName?: string): boolean => {
+          const errorMessage = error instanceof Error ? error.message : String(error);
+
+          // Gemini CLI: Project ID required for workspace accounts
+          if (errorMessage.includes('GOOGLE_CLOUD_PROJECT') || errorMessage.includes('GOOGLE_CLOUD_PROJECT_ID')) {
```
wladimiiir: We need to separate the providers from the `agent.ts` file. Any provider-specific logic should be part of the `LlmProviderStrategy`, and the `ModelManager` should include a function to be called by the `Agent` class. In this case, it could be something like `modelManager.isRetryable(provider, model, ...)`, using positive rather than negative naming.

src/main/models/providers/gemini-cli.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  // Dynamic import required: ai-sdk-provider-gemini-cli is an ESM module
+  // and cannot be loaded synchronously in Electron's CommonJS environment
+  const { createGeminiProvider } = await import('ai-sdk-provider-gemini-cli');
+
+  // Set GOOGLE_CLOUD_PROJECT env var if projectId is configured (for organization/enterprise accounts)
+  // Note: The underlying @google/gemini-cli-core library only reads projectId from process.env,
+  // so we must set it here. This is a limitation of the library, not our implementation.
+  if (projectId) {
+    process.env.GOOGLE_CLOUD_PROJECT = projectId;
```
wladimiiir, line 96: I could not find any mention of this environment variable. Is this something really supported?

jutaz:
> Left some comments. The most important is to move the isErrorRetryable into ModelManager class and the LlmProviderStrategy to keep the separation of concerns.

yep, good points, addressing 👍

src/common/agent.ts
jutaz: `gemini-3-pro-preview ` is still in preview (not everyone has access to it), so wanted to default to the same model cli defaults to, which is basically `gemini-2.5-flash` currently.

src/main/models/providers/gemini-cli.ts
jutaz: yep: https://geminicli.com/docs/get-started/authentication/#use-vertex-ai

PR #675, opened 2026-02-20
Bugfix - Improve process killing reliability

jutaz:
I tend to run a fair amount of commands per session, and I think this _may_ be one of the sources of UI slowdown after a while - some of my commands time out, having the command runner retry them, but they tend to linger on for a while and sometimes I do notice orphaned processes after timeouts - this may be the issue.

This PR aims to ensure that any child process started via power bash tool is properly and fully killed as soon as the context is closed (either timeout or thread is stopped, for example).


src/main/agent/tools/power.ts
```diff
@@ -573,31 +592,50 @@ Do not use escape characters \\ in the string like \\n or \\" and others. Do not
...
+            stderr = 'Operation was cancelled by user.';
+            exitCode = 130; // Standard cancel exit code (128 + SIGINT=2)
+            resolveWithResult();
+          };
+
+          // Add abort listener only if abortSignal and childProcess exist
+          if (abortSignal && childProcess) {
+            if (abortSignal.aborted) {
+              // Signal already aborted, call handler synchronously
+              await abortListener();
```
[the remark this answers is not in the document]
jutaz: not true. Comment is wrong tho

src/main/agent/tools/power.ts
```diff
@@ -624,32 +643,50 @@ Do not use escape characters \\ in the string like \\n or \\" and others. Do not
...
-
-        // Listen for abort signal
-        abortSignal?.addEventListener('abort', abortListener);
-
         try {
           const childProcess = spawn(command, {
             cwd: absoluteCwd,
             shell: true,
-            env: process.env,
+            env: { ...process.env, TERM: 'dumb', DEBIAN_FRONTEND: 'noninteractive' },
```
wladimiiir: Do we need those? It seems like you have added some local specific env variables.

jutaz: Not quite - these are to make sure commands aren't interactive - this should improve the robustness of the commands too since it won't hang around waiting for an input.

PR #678, opened 2026-02-21
Bugfix - Subagents do not include base rules

jutaz:
As per my investigation, subagents do not receive AGENTS.md and `.aider-desk/rules/` files when they have a custom `systemPrompt` configured. The bug was in `agent.ts` where `getSystemPrompt()` was only called when `systemPrompt` was undefined - but subagents pass their custom prompt, causing the base prompt with rules to be skipped entirely.

### DIFF
--- a/resources/prompts/rules-block.hbs
+++ b/resources/prompts/rules-block.hbs
@@ -1,0 +1,5 @@
+<Knowledge>
+  <Rules>
+{{{rulesFiles}}}
+  </Rules>
+</Knowledge>
--- a/src/common/agent.ts
+++ b/src/common/agent.ts
@@ -424,6 +424,7 @@ export const DEFAULT_AGENT_PROFILE: AgentProfile = {
     color: '#3368a8',
     description: '',
     contextMemory: ContextMemoryMode.Off,
+    includeRules: false,
   },
   ruleFiles: [],
 };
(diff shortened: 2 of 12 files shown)


aresstokrat:
Just wanted to confirm this bug from a real-world multi-agent setup.

I'm running a project with Architect (main agent) delegating to Scout (GLM-4.5-air) and Builder subagents. I have project hygiene rules configured as Skills with checkboxes enabled for both agents. Despite this, subagents consistently ignore file placement rules — creating temp scripts, diagnostic files, and test artifacts directly in the project root.

Debugging the session logs revealed the root cause matches exactly what this PR describes: the Scout subagent received only ~2,100 tokens in its context — just the bare task prompt with zero rules or skills attached. The Architect sees everything correctly, but the moment it delegates via `run_task`, all project rules vanish.

This has been a significant pain point — the project accumulated dozens of misplaced files (test_*.js, check_*.py, temp_*.ts, fix_*.js scattered across root directories) because subagents had no awareness of the file structure conventions.

Great catch @jutaz — looking forward to this being merged.

aresstokrat:
One thought for a potential follow-up (not for this PR): now that base rules are correctly propagated, it might be worth considering per-profile rule filtering for subagents in the future.

In my setup, the Architect profile has extensive rules (deploy runbook, db-safety, release protocol, methodology docs) that are irrelevant for a Scout subagent doing read-only file navigation, and they eat into the attention budget of smaller models like GLM-4.5-air.

Something like a `rules` or `includeRules` field in subagent profile config — so each subagent only receives the rules relevant to its role — would be a nice optimization. But that's a separate feature, this bugfix is the right first step.

wladimiiir:
@aresstokrat 
It is possible to add rule files to a specific agent, by adding it into `.aider-desk/agents/my-agent/rules` folder. So in case you have a agent that's specifically used only as a subagent, you can put the rules there.

PR #694, opened 2026-03-07
fix: allow reverting untracked (new) files in updated files sidebar

jutaz:
This PR fixes an issue where clicking "Revert changes" on a new (untracked) file in the updated files sidebar would silently fail because  only works on tracked files.

The fix:
1. Detects if a file is untracked using git.
2. If tracked: Uses  (existing behavior)
3. If untracked: Deletes the file using  (files/symlinks) or  (directories)
4. Uses absolute paths for filesystem operations (no  support in Node fs functions)
5. Refined error handling to only treat specific git errors as "untracked"

Tests have been added to cover tracked files, untracked files, symlinks, directories, and various error cases.

### DIFF
--- a/src/main/worktrees/worktree-manager.ts
+++ b/src/main/worktrees/worktree-manager.ts
@@ -1528,7 +1528,7 @@ export class WorktreeManager {
         isTracked = true;
       } catch (error) {
         const errorMessage = error instanceof Error ? error.message : String(error);
-        const isUntrackedFileError = errorMessage.includes('did not match') || errorMessage.includes('did not match any file');
+        const isUntrackedFileError = errorMessage.includes('did not match');
 
         if (isUntrackedFileError) {
           isTracked = false;


wladimiiir:
@jutaz How do you get the files visible in the Updated Files section? It should show only tracked files.

jutaz:
> @jutaz How do you get the files visible in the Updated Files section? It should show only tracked files.

Great question - and I don't have a good answer for you - seems that I just see them. My (relevant parts of) `.gitconfig`:
```
[column]
  ui = auto
[branch]
  sort = -committerdate
[tag]
  sort = version:refname
[init]
  defaultBranch = main
[diff]
  algorithm = histogram
  colorMoved = plain
  mnemonicPrefix = true
  renames = true
[push]
  default = simple
  autoSetupRemote = true
  followTags = true
[fetch]
  prune = true
  pruneTags = true
  all = true

# why the hell not?

[help]
  autocorrect = prompt
[commit]
  verbose = true
[rerere]
  enabled = true
  autoupdate = true
[rebase]
  autoSquash = true
  autoStash = true
  updateRefs = true
```

PR #781, opened 2026-05-11
Fix add-file handling for existing context files.

JsonSong89:
Fix add-file handling for existing context files.

Previously, `ContextManager.addContextFile` returned an empty array whenever the resolved file path was already present in the task context. This caused `Task.addFiles` to treat the operation as a no-op, so an existing readonly context file could not be updated when a later `add-file` message provided a different `readOnly` value.

The fix changes `addContextFile` to update the existing context file metadata when the file is already present but the incoming `readOnly` value differs. The method signature remains unchanged and it still returns the list of changed files. This allows `Task.addFiles` to reuse its existing update flow and notify connectors/UI/context-info when the readonly state changes.

### DIFF
--- a/src/main/task/context-manager.ts
+++ b/src/main/task/context-manager.ts
@@ -110,7 +110,15 @@ export class ContextManager {
       return [];
     }
 
-    if (this.isContextFileAlreadyAdded(absolutePath)) {
+    const existingContextFile = this.findExistingContextFile(absolutePath);
+    if (existingContextFile) {
+      const readOnly = contextFile.readOnly ?? false;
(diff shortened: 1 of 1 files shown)


src/main/task/context-manager.ts
```diff
@@ -177,8 +185,8 @@ export class ContextManager {
    * Checks whether a resolved absolute path matches any already-added context file.
```
wladimiiir: Comment should be updated to reflect the new version of the function.

wladimiiir (approved):
Thank you for your contribution.

PR #879, opened 2026-07-25
Fix: include additionalPaths in getShellPath() PATH resolution

marceli1404:
In \getShellPath()\, \dditionalPaths\ (npm global bins, yarn bins, nvm node versions, Git paths on Windows, common Linux paths) is computed but never included in the \combinedPaths\ Set that becomes the cached PATH.

Line 513 was missing \...additionalPaths\:\\\
- const combinedPaths = new Set([...shellPath.split(pathSep), ...currentPath.split(pathSep)]);
+ const combinedPaths = new Set([...shellPath.split(pathSep), ...currentPath.split(pathSep), ...additionalPaths]);
\\\

Without this fix, npm/yarn/nvm-installed tools and Git commands may not be found on systems where they aren't already in the base PATH.

### DIFF
--- a/src/main/logger/__tests__/log-buffer.test.ts
+++ b/src/main/logger/__tests__/log-buffer.test.ts
@@ -1,0 +1,28 @@
+import { describe, expect, it } from 'vitest';
+import { SystemLogEntry } from '@common/types';
+import { LogBuffer } from '../log-buffer';
+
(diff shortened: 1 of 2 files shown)


wladimiiir:
Thanks for the fix. Can you please fix the lint issues?

PR #885, opened 2026-07-26
fix(aider): clear stale state after natural process exit

ag-linden:
## Summary

A naturally exited Aider child process remained stored in `AiderManager`, so `isStarted()` stayed true and later commands could not restart Aider. The close handler now clears the process reference, PID file, command output state, and startup state, rejects a pending startup wait when the connector never became ready, and ignores stale close events from a replaced process.

## Verification

- `npm run test:node -- src/main/task/__tests__/aider-manager.test.ts` — passed
- `npm run typecheck` — passed
- `npm exec eslint -- src/main/task/aider-manager.ts src/main/task/__tests__/aider-manager.test.ts` — passed
- `git diff --check` — passed

The change is limited to the managed Aider child-process lifecycle; no data migration or API change is involved.

### DIFF
--- a/src/main/task/__tests__/aider-manager.test.ts
+++ b/src/main/task/__tests__/aider-manager.test.ts
@@ -1,0 +1,38 @@
+import { describe, expect, it } from 'vitest';
+
+import { AiderManager } from '../aider-manager';
+
+type FakeAiderProcess = { pid?: number };
+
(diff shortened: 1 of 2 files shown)


wladimiiir (approved):
The changes look good, thank you for your contribution.

PR #912, opened 2026-08-24
refactor(worktrees): prune stale worktrees and reorder creation steps

sbodese:
Fix with Test about the reported issue:

Hi Hotovoteam: i mean the worktree Feature is partially buggy, i created a project and let create a task list, and if i want to implement a task, i switch to the worktree mode, but AiderDesk says, "no" and i see this error in the logs:

genericNodeError (node:internal/errors:985:15)\n    at wrappedFn (node:internal/errors:539:14)\n    at ChildProcess.exithandler (node:child_process:417:12)\n    at ChildProcess.emit (node:events:509:28)\n    at maybeClose (node:internal/child_process:1124:16)\n    at ChildProcess._handle.onexit (node:internal/child_process:306:5)","stderr":"Preparing worktree (checking out 'aider-desk/task/docsai-t2-config-model-validation')\nfatal: 'C:/work/DocsAI/.aider-desk/tasks/19321b6c/worktree' is a missing but already registered worktree;\nuse 'add -f' to override, or 'prune' or 'remove' to clear\n","stdout":"","timestamp":"2026-08-24T11:09:09.082Z"}

**Further Analysis:**
I
Root cause: a task's workingMode is fixed at creation time — there's no way to switch an existing task to worktree mode. Your prepared T1–T10 series was created as local tasks, so AiderDesk never gives them real git worktrees. Evidence from the repo:

git worktree list shows real worktrees only for the tasks created with worktree mode (b25619dc = this task, e79df709 = plan task)
T1's worktree dir is a plain file copy (no .git), T2's is an empty leftover (marked prunable) — AiderDesk won't use them
Proposed plan:

Re-create the 9 prepared tasks (T2–T10) as new tasks with worktree=true, same names, same agent profile (dotnet), copying each original initial prompt verbatim so nothing is lost.
Delete the 9 old local-mode tasks (T2–T10) to avoid duplicate series. (T1 stays — it's READY_FOR_REVIEW and its scaffold is already committed on master.)
Cleanup: git worktree prune to drop the stale T2 worktree registration and empty dir.
Then you work the series one at a time as before; each task gets its own branch (aider-desk/task/...) and you merge sequentially so the "assume task N-1 done" prompts hold.
Side note: the old memory claiming these were plain subdirs is outdated — worktree-mode tasks here are real git worktrees; I'll correct that at the end.

### DIFF
--- a/src/main/worktrees/__tests__/worktree-manager.create.test.ts
+++ b/src/main/worktrees/__tests__/worktree-manager.create.test.ts
@@ -1,0 +1,128 @@
+import { join } from 'path';
+
+import { beforeEach, describe, expect, it, vi, type Mock } from 'vitest';
+
+import { WorktreeManager } from '../worktree-manager';
+
+vi.mock('@/logger', () => ({
+  default: {
+    info: vi.fn(),
+    warn: vi.fn(),
+    error: vi.fn(),
+    debug: vi.fn(),
+  },
+}));
+
+vi.mock('@/utils', () => ({
+  execWithShellPath: vi.fn(),
+  withLock: vi.fn((_id: string, fn: () => Promise<unknown>) => fn()),
+}));
+
+vi.mock('fs/promises', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('fs/promises')>();
+  return {
+    ...actual,
+    rm: vi.fn(),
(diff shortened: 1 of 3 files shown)


wladimiiir (approved):
Thank you

PR #916, opened 2026-08-27
fix: worktree merge failing on in-progress rebase and untracked stash…

sbodese:
… files

Root cause (from the 2026-08-26 error log of v0.80.0 on a multi-worktree project):

1. Squash-merging a worktree while a rebase with unresolved conflicts was still in progress started a new "git rebase --onto", which failed with "already a rebase-merge directory". The failure path then ran "git rebase --abort" - silently destroying the rebase conflict state the user was in the middle of resolving.

2. When a merge had to be rolled back, the main branch's stashed uncommitted changes were restored with "git stash apply". A previous partial apply had already brought the stash's untracked files back to disk (e.g. a file that the merged commit introduced as tracked), so the apply failed with "already exists, no checkout". "git reset --hard" cannot remove untracked files, so the retry failed again and the user's uncommitted master changes were left trapped in a leftover stash entry.

Fixes in WorktreeManager:

- Guard squash merge, plain merge and rebase against a rebase that is already in progress (clear error message instead of a git fatal).
- Only abort a rebase that the failing operation started itself: the abort is skipped when git reports a pre-existing rebase, so a user's in-progress conflict rebase is never destroyed.
- "git stash apply" now retries after an untracked-file collision: files that are untracked in the index and content-identical to the copy in the stash (verified via git hashes) are removed first; tracked files and files with different content are never touched.
- Merge failure recovery only runs "git merge --abort" / "git reset --hard" when the merge step actually ran, so a failed stash step can no longer destroy uncommitted changes in the main repo.
- Specific GitErrors are no longer re-wrapped into a generic message, and a finally-failed main-branch stash recovery reports the stash ID so the changes can be recovered manually via "git stash list".

Adds unit tests for the rebase guards, the conditional rebase abort and the stash untracked-collision handling, and updates the rebase-validation mocks for the new guard.


wladimiiir (approved):
Thank you

PR #918, opened 2026-08-28
fix: strip image input for non-vision models

sbodese:
Local models (e.g. Ollama) that lack vision support began failing with 'image input is not supported' after the v0.80.0 image-extraction feature. Vision capability is now detected per model and image parts are stripped before a model call when the target is non-vision.

Changes:

- Adds Model.supportsVision; Ollama vision detection via /api/show (defaults to false on error)
- Adds AgentProfile.sendImagesToModel with override logic (undefined = use global default, explicit true/false = per-profile override) — follows the same pattern as temperature and maxTokens
- ModelManager.modelSupportsVision treats unknown as vision-capable
- Adds stripImageParts applied in optimizeMessages when the resolved target is non-vision
- Per-profile "Send Images to Model" checkbox in ProfileGeneralSection with clear-override button
- Bumps provider-models cache version; adds tests


src/main/agent/agent.ts
```diff
@@ -920,6 +923,7 @@ export class Agent {
         };
 
         const optimizedMessages = await getOptimizedMessages();
```
wladimiiir, line 926: It would be better to have the `includeImageParts` argument on `getOptimizedMessages` function and pass it to it, so it would not include it in the first place if not needed.

src/renderer/src/components/settings/agent/AgentSettings.tsx
```diff
@@ -366,6 +368,10 @@ export const AgentSettings = ({
     setSettings(updatedSettings);
   };
 
+  const handleSendImagesToModelChange = (checked: boolean) => {
+    setSettings({ ...settings, sendImagesToModel: checked });
```
wladimiiir: This is storing the setting on global level, but for this context it would make more sense to have it on `AgentProfile` level, so each agent can decide.

src/renderer/src/components/settings/agent/AgentSettings.tsx
```diff
@@ -599,6 +605,22 @@ export const AgentSettings = ({
             <span className="text-sm font-medium uppercase truncate">{selectedProfile?.name || t('settings.agent.selectOrCreateProfile')}</span>
           </div>
         )}
+
+        <div className="px-6 pt-4 pb-2 border-b border-border-default flex-shrink-0">
+          <div className="max-w-3xl mx-auto">
+            <Checkbox
```
wladimiiir: This should be moved to `ProfileGeneralSection` and it should be follow the "override" logic as for other parameters from a model (Temperature, Max Tokens...).

wladimiiir (changes requested):

src/main/agent/agent.ts
sbodese: i check it ...

src/renderer/src/components/settings/agent/AgentSettings.tsx
sbodese: i check it ...

src/renderer/src/components/settings/agent/AgentSettings.tsx
sbodese: i check it ...

src/main/agent/agent.ts
sbodese: you mean shouldSendImages? i dont use a includeImageParts literal/variable ... Is this a human test here ^^?

sbodese: oki, moved

src/renderer/src/components/settings/agent/AgentSettings.tsx
sbodese: done

src/renderer/src/components/settings/agent/AgentSettings.tsx
sbodese: done

src/main/agent/tools/tasks.ts
```diff
@@ -351,15 +350,16 @@ export const createTasksToolset = (settings: SettingsData, task: Task, profile:
         .optional()
         .default(false)
         .describe('If true, the task will be created in worktree mode. Use only when explicitly requested by the user.'),
-      executeAndWait: z
-        .boolean()
+      executionMode: z
+        .enum(TaskExecutionMode)
```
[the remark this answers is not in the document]
sbodese: it seems that this bot recommends a deprecated nativeEnum api ?!

wladimiiir (changes requested):
@sbodese 
Sorry for the late reply. There are a few findings I would like addressed before merging the changes. Thanks.

1. **Global `sendImagesToModel` has no UI** — the `SettingsData` field is honored in the fallback chain, but no renderer setting exposes it; users can only toggle per-profile (and the profile "Override" button only sets `true`). Either add UI or drop the field.
2. **`estimateTokens` (agent.ts:1762) doesn't pass `shouldSendImages`** → token estimates include images that would be stripped (inflated for non-vision models).
3. **Conversation compaction** (`Agent.generateText` :1648 via `compactConversation`) and `generateObject` don't strip images — edge case if the auxiliary model is non-vision while history contains images.
4. **Ollama fail-closed asymmetry**: `/api/show` error/timeout → `supportsVision: false`, which gets disk-cached; a transiently slow Ollama server silently disables image input until next reload. Contradicts the "unknown = assume vision" principle; consider returning `undefined` on error instead (the test would need updating too).
5. Cosmetic: ModelDialog shows detected Yes/No rather than "Auto" for Ollama models; saving any edit persists detection as an explicit override.

PR #919, opened 2026-08-28
fix: auto-compact for models with unknown context size

sbodese:
Decouple the auto-compact token threshold from maxInputTokens so compaction still triggers when a model's context size is unknown, preventing prompts from growing past the model's hard limit. Also read Ollama models' maxInputTokens from the reported context_length so the percentage-based threshold and token bar work for Ollama (bumps provider-models cache version).

based on this error log, i have configured a context window of 240000, so this parameter have no impact

image

image

responseBody":"{"error":{"message":"This model's maximum context length is 262144 tokens. However, you requested 0 output tokens and your prompt contains at least 262145 input tokens, for a total of at least 262145 tokens. Please reduce the length of the input prompt or the number of requested output tokens. (parameter=input_tokens, value=262145)","type":"BadRequestError","param":"input_tokens","code":400}}"

### DIFF
--- a/src/main/agent/__tests__/agent.test.ts
+++ b/src/main/agent/__tests__/agent.test.ts
@@ -28,7 +28,7 @@ import path from 'path';
 // @ts-expect-error istextorbinary is not typed properly
 import { isBinary } from 'istextorbinary';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
-import { ToolApprovalState } from '@common/types';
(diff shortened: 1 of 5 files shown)


sbodese:
@wladimiiir maybe its needs more a manual update because this switch here "works", but here also this "max tokens" setting

image

also my question: what is the relation to the "context window"?

wladimiiir (approved):
Thanks.

PR #930, opened 2026-09-05
fix: atomic context save and corrupt recovery

pcfreak30:
Writes task context atomically (temp file + rename) serialized through `withLock`, and recovers from a corrupt or truncated `context.json` by falling back to the newest backup or an empty context.

Adds tests for atomic save and corrupt-context recovery.


wladimiiir (approved):
Thanks.

PR #931, opened 2026-09-05
fix(extensions): log tps-counter diagnostics at debug level

pcfreak30:
The TPS counter extension logged message UUIDs at info level every time `getUIExtensionData` was polled, flooding the logs during normal use.

All diagnostic logging (getUIExtensionData key dumps, onResponseCompleted traces, message data key dumps) now uses debug level, so it only appears when debug logging is enabled. Info level is kept for the extension-loaded message and the user-facing toggle confirmations.


wladimiiir (changes requested):
Version also needs to be increased to `1.1.1`

wladimiiir (approved):

PR #932, opened 2026-09-05
feat(openai-codex): dedicated sign-in ui and device flow

pcfreak30:
Sign-in becomes an explicit settings action instead of a side effect inside the model-creation path, and the OpenAI Codex provider gains a device-code flow so browsers on a different machine than AiderDesk (server/headless) can authenticate.

- Adds device code sign-in (with polling, cancel, and feature-disabled errors) alongside the loopback browser flow, which now binds an ephemeral port and derives the redirect URI from the actual port
- Moves auth out of the prompt path: the provider only loads stored tokens and fails with a clear sign-in error when unauthenticated
- Stores tokens in the AiderDesk data dir (0600, survives extension updates) and surfaces real network errors instead of bare fetch failures
- Adds an extension settings panel with auth status, sign-in/sign-out, and a 15s timeout plus request dedup on usage/auth calls

### DIFF
--- a/packages/extensions/extensions/openai-codex/ConfigComponent.jsx
+++ b/packages/extensions/extensions/openai-codex/ConfigComponent.jsx
@@ -78,11 +78,11 @@
 
       {isPending && (
         <div className="flex flex-col gap-3 rounded-lg border border-border-primary p-4">
-          {config.pendingFlow.type === 'device' && deviceCode ? (
+          {config.pendingFlow?.type === 'device' && deviceCode ? (
             <>
               <p>1. Open the sign-in page in any browser:</p>
-              <a href={config.pendingFlow.verificationUrl} target="_blank" rel="noopener noreferrer" className="text-left text-blue-500 underline">
-                {config.pendingFlow.verificationUrl}
+              <a href={config.pendingFlow?.verificationUrl} target="_blank" rel="noopener noreferrer" className="text-left text-blue-500 underline">
+                {config.pendingFlow?.verificationUrl}
               </a>
               <p>2. Enter this one-time code:</p>
               <div className="rounded bg-bg-tertiary px-4 py-2 font-mono text-xl tracking-widest">{deviceCode}</div>
--- a/packages/extensions/extensions/openai-codex/ConfigComponent.jsx
+++ b/packages/extensions/extensions/openai-codex/ConfigComponent.jsx
@@ -1,0 +1,103 @@
+({ config, updateConfig, executeExtensionAction, ui }) => {
+  const { useEffect } = React;
+  const { Button } = ui;
+
+  const handleAction = async (action) => {
+    const result = await executeExtensionAction(action);
+    if (result) {
+      updateConfig(result);
+    }
+  };
+
+  const isPending = !!config?.pendingFlow;
+  const lastError = config?.lastError;
+
+  useEffect(() => {
+    if (!isPending) {
+      return undefined;
+    }
+    const interval = setInterval(async () => {
+      try {
+        const state = await executeExtensionAction('getAuthState');
+        if (state) {
+          updateConfig(state);
+        }
+      } catch {
+        // dialog still open while main process may be busy — retry on next tick
+      }
+    }, 3000);
+    return () => clearInterval(interval);
+  }, [isPending, executeExtensionAction, updateConfig]);
+
+  const deviceCode = config?.pendingFlow?.userCode;
+
+  return (
+    <div className="flex flex-col gap-4 text-sm">
+      {config?.status === 'signed-in' && (
+        <div className="flex flex-col gap-4">
+          <div className="flex items-center gap-2">
+            <span className="text-green-500">&#10003;</span>
+            <span>
+              Signed in as{' '}
+              <span className="font-medium">{config?.email ? config.email : config?.accountId ? `account ${config.accountId.slice(0, 8)}…` : 'ChatGPT account'}</span>
+            </span>
+          </div>
+          <Button variant="outline" onClick={() => handleAction('signOut')} disabled={isPending} className="w-full">
+            Sign out
+          </Button>
+        </div>
+      )}
+
+      {config?.status === 'expired' && !isPending && <div className="text-yellow-500">Session expired — sign in again to refresh it.</div>}
+
+      {!isPending && config?.status !== 'signed-in' && (
+        <div className="flex flex-col gap-3">
+          <p className="text-text-secondary">Sign in with your ChatGPT Plus/Pro account. No API key required.</p>
+
+          <div className="flex items-center gap-2">
+            <Button onClick={() => handleAction('signInBrowser')} disabled={isPending}>
+              Sign in with Browser
+            </Button>
+            <Button variant="outline" onClick={() => handleAction('signInDeviceCode')} disabled={isPending}>
+              Sign in with Device Code
+            </Button>
+          </div>
+          <p className="text-xs text-text-secondary">
+            Use <span className="font-medium">Device Code</span> when your browser runs on a different machine than AiderDesk (e.g. remote or headless
+            server). Device code sign-in must be enabled in your ChatGPT security settings.
+          </p>
+        </div>
+      )}
+
+      {lastError && (
+        <div className="rounded-lg border border-red-500/40 p-3 text-red-500">
+          {lastError}
+          {!isPending && <div className="mt-1 text-xs text-text-secondary">State has been reset — you can retry sign-in above.</div>}
+        </div>
+      )}
+
+      {isPending && (
+        <div className="flex flex-col gap-3 rounded-lg border border-border-primary p-4">
+          {config.pendingFlow.type === 'device' && deviceCode ? (
+            <>
+              <p>1. Open the sign-in page in any browser:</p>
+              <a href={config.pendingFlow.verificationUrl} target="_blank" rel="noopener noreferrer" className="text-left text-blue-500 underline">
+                {config.pendingFlow.verificationUrl}
+              </a>
+              <p>2. Enter this one-time code:</p>
+              <div className="rounded bg-bg-tertiary px-4 py-2 font-mono text-xl tracking-widest">{deviceCode}</div>
+              <p className="text-xs text-text-secondary">Waiting for approval… this expires in 15 minutes.</p>
+            </>
+          ) : config.pendingFlow.type === 'device' ? (
+            <p>Waiting for the device code…</p>
+          ) : (
+            <p>Waiting for sign-in to complete in your browser…</p>
+          )}
+          <Button variant="outline" size="sm" className="self-start" onClick={() => handleAction('cancelSignIn')}>
+            Cancel
+          </Button>
+        </div>
+      )}
+    </div>
+  );
+}
--- a/packages/extensions/extensions/openai-codex/index.ts
+++ b/packages/extensions/extensions/openai-codex/index.ts
@@ -1,10 +1,11 @@
 import { randomBytes, createHash } from 'node:crypto';
 import { createServer, type Server } from 'node:http';
-import { readFile, writeFile, mkdir } from 'node:fs/promises';
-import { readFileSync } from 'node:fs';
+import type { AddressInfo } from 'node:net';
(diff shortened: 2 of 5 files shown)


packages/extensions/extensions/openai-codex/index.ts
```diff
@@ -213,49 +246,89 @@ const getAccountId = (accessToken: string): string | null => {
...
+        grant_type: 'authorization_code',
+        client_id: getClientId(),
+        code,
+        code_verifier: codeVerifier,
+        redirect_uri: redirectUri,
+      }),
+      signal: AbortSignal.timeout(FETCH_TIMEOUT_MS),
+    });
+  } catch (error) {
+    throw networkError(error, TOKEN_URL);
```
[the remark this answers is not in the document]
pcfreak30: \@kilocode-bot False positive. `networkError` is referenced at line 295 *inside the body of `exchangeAuthorizationCode`*, which only executes at runtime — long after module evaluation has initialized the `const networkError` binding. TDZ violations only occur when a `const`/`let` binding is *executed* before initialization; defining a function earlier in the module that closes over it is the standard pattern (same as the other helpers here). No runtime error is possible, so no change.

packages/extensions/extensions/openai-codex/ConfigComponent.jsx
```diff
@@ -0,0 +1,103 @@
...
+      {lastError && (
+        <div className="rounded-lg border border-red-500/40 p-3 text-red-500">
+          {lastError}
+          {!isPending && <div className="mt-1 text-xs text-text-secondary">State has been reset — you can retry sign-in above.</div>}
+        </div>
+      )}
+
+      {isPending && (
+        <div className="flex flex-col gap-3 rounded-lg border border-border-primary p-4">
+          {config.pendingFlow.type === 'device' && deviceCode ? (
```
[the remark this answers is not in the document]
pcfreak30: \@kilocode-bot Hardened in 1a7d1059 — the `config.pendingFlow` accesses inside the conditional now use optional chaining (`?.type`, `?.verificationUrl`), so a config race can no longer throw here.

packages/extensions/extensions/openai-codex/index.ts
```diff
@@ -310,165 +390,309 @@ const startOAuthServer = (expectedState: string): Promise<{ server: Server; wait
...
+    deviceFlow.abort.abort();
+    deviceFlow = null;
   }
 };
 
-// --- Get valid access token ---
+// Only mutates the flow that failed — a late error from an aborted flow must not taint a newer one
+const failDeviceFlow = (message: string, context: ExtensionContext): void => {
+  context.log(`Device code sign-in failed: ${message}`, 'warn');
+  lastAuthError = message;
```
wladimiiir, line 531: **BLOCKING: this violates the invariant stated in the comment above** ("a late error from an aborted flow must not taint a newer one").

`failDeviceFlow` takes no flow identity: it sets `lastAuthError` unconditionally and aborts/nulls whatever `deviceFlow` *currently* is — which may be a newer flow than the one that failed.

Concrete race: flow A's poll rejects with a non-abort network error just after the user cancels A and starts flow B → this function aborts B (B's `sleep` rejects with `AbortError`, the catch at the bottom swallows it, B ends with no state change) and shows A's error as if B failed. The user's active sign-in silently dies.

The `deviceFlow !== flow` guard pattern already exists above (before mutating `flow.userCode` and after success). Suggested fix — pass the failed flow in and early-return when it is no longer current:

```ts
const failDeviceFlow = (flow: DeviceFlowState, message: string, context: ExtensionContext): void => {
  if (deviceFlow !== flow) return;
  context.log(`Device code sign-in failed: ${message}`, 'warn');
  lastAuthError = message;
  deviceFlow.abort.abort();
  deviceFlow = null;
};
```

packages/extensions/extensions/openai-codex/index.ts
```diff
@@ -561,11 +810,83 @@ export default class OpenAICodexAuthExtension implements Extension {
...
+          }
+        })();
+        return getAuthState();
+      }
+
+      case 'signInDeviceCode':
+        startDeviceCodeFlow(context);
+        return getAuthState();
+
+      case 'cancelSignIn':
```
wladimiiir, line 919: **WARNING: `cancelSignIn` does not actually cancel the browser flow.**

`runBrowserOAuthFlow` has no AbortController, so this only clears `browserFlowRunning` — the loopback server keeps listening for up to 10 minutes and the OAuth flow keeps running. Consequences:

1. Completing the login in the browser *after* cancel (or after Sign out) still saves tokens.
2. After cancel, the user can start a second browser flow; when the stale flow times out, `lastAuthError` is set unconditionally in the `signInBrowser` catch (line 848, no flow-identity guard) — a false "Sign-in timed out" error can appear after a successful sign-in, and the cleared guard then even allows a third concurrent flow.

Suggestion: give the browser flow an AbortController + identity guard mirroring the device flow, and make cancel actually abort it.

packages/extensions/extensions/openai-codex/index.ts
```diff
@@ -71,13 +82,19 @@ let currentSessionId: string | undefined;
 const clampCacheKey = (key: string): string =>
   key.length <= PROMPT_CACHE_KEY_MAX_LENGTH ? key : Array.from(key).slice(0, PROMPT_CACHE_KEY_MAX_LENGTH).join('');
 
-// Token storage
-const TOKEN_FILE = join(__dirname, 'auth-token.json');
+// Tokens live in the app data dir so they survive extension updates; the legacy location inside
+// the extension install dir is migrated on load.
+const DATA_DIR = process.env.AIDER_DESK_DATA_DIR
+  ? join(process.env.AIDER_DESK_DATA_DIR, 'extensions-data', 'openai-codex')
+  : join(homedir(), '.aider-desk', 'extensions-data', 'openai-codex');
```
wladimiiir: **WARNING: this path does not match the app's data directory and ignores the home-dir override.**

`AIDER_DESK_DATA_DIR` is only *read* in `src/main/paths.ts` — the app never sets it — so on default installs this fallback is used. But `~/.aider-desk` is neither Electron `userData` (what `AIDER_DESK_DATA_DIR` in `src/main/constants.ts:8` actually resolves to) nor the env-paths dir used in server mode. It does satisfy "survives extension updates", but:

- The README/PR claim "tokens are stored in the AiderDesk data directory" is inaccurate on default installs.
- Custom `AIDER_DESK_HOME_DIR` installs (`src/main/constants.ts:24`) are ignored — tokens still land in `~/.aider-desk` even when the app home was moved.
- `extensions-data` is a new directory convention invented by this extension rather than a framework-provided location.

Suggestion: at minimum derive from `process.env.AIDER_DESK_HOME_DIR ?? join(homedir(), '.aider-desk')`, or better, expose a data dir on `ExtensionContext` so extensions don't each guess paths.

packages/extensions/extensions/openai-codex/index.ts
```diff
@@ -561,11 +810,83 @@ export default class OpenAICodexAuthExtension implements Extension {
...
+
   getProviders(context: ExtensionContext): ProviderDefinition[] {
     const createLlm = async (_profile: ProviderProfile, model: Model) => {
-      context.log(`Creating OpenAI Codex model: ${model.id}`, 'info');
-
-      const { accessToken, accountId } = await getValidAccessToken(context);
+      // Never start OAuth here — sign-in is an explicit user action from the settings UI
+      const tokens = await getValidTokensNonInteractive(context);
+      if (!tokens) {
+        throw new Error('OpenAI Codex is not authenticated. Open extension settings for OpenAI Codex Auth and sign in.');
```
wladimiiir: **WARNING: this error conflates "not authenticated" with a transient refresh failure.**

`getValidTokensNonInteractive` returns `null` both when no tokens exist *and* when a refresh attempt fails (network blip, the new 15s timeout, OpenAI 5xx). The message then tells the user to open settings and sign in — which won't fix a network issue and is unnecessary if tokens exist. This is also a behavior regression vs the old flow, which fell back to automatic re-auth on refresh failure.

Suggestion: distinguish the two cases (e.g., a typed result or separate error messages for "no tokens" vs "token refresh failed: <cause>") so the error is actionable.

packages/extensions/extensions/openai-codex/index.ts
```diff
@@ -310,165 +390,309 @@ const startOAuthServer = (expectedState: string): Promise<{ server: Server; wait
...
+        }
+        lastAuthError = undefined;
+        quotaCache.data = null;
+        quotaCache.lastFetchTime = 0;
+        context.triggerUIDataRefresh(STATUS_BAR_COMPONENT_ID);
+        return;
+      }
+
+      // 403/404 mean "not approved yet" — keep polling until the deadline
+      if (pollResponse.status !== 403 && pollResponse.status !== 404) {
```
wladimiiir: **WARNING: the polling contract rests on unverifiable status-code assumptions.**

Treating only 403/404 as "pending" assumes OpenAI's undocumented `/api/accounts/deviceauth/*` endpoints signal `authorization_pending` that way. Standard device grants instead return 400 with an error code, and `slow_down` requires increasing the poll interval — neither is handled here, so an unexpected pending/slow-down response aborts the whole flow or hammers the endpoint every `intervalMs`.

Relatedly, line 535 maps *any* 404 from the usercode endpoint to "Device code sign-in is not enabled for this account" — a generic route 404 (API change) would send users into account settings on a wild goose chase.

Suggestion: reference the codex-rs implementation this mirrors in a comment, and map statuses conservatively (e.g., also treat 400-with-pending-code as pending, back off on slow_down).

wladimiiir (commented):
Review findings: 1 blocking, 4 warnings (test coverage discussed separately, not included here). Traced runtime flows for model creation when unauthenticated, browser vs device sign-in, cancel/sign-out, and quota/auth dedup against the extension framework contract (ExtensionSettingsDialog, extension-manager install logic, transpiler).

packages/extensions/extensions/openai-codex/index.ts
pcfreak30: Fixed in f84c5bc3 — `failDeviceFlow` now takes the failing `DeviceFlowState` and early-returns when it is no longer the current flow, so a late error from flow A can neither abort nor taint flow B (exactly the guard pattern you suggested).

packages/extensions/extensions/openai-codex/index.ts
pcfreak30: Fixed in f84c5bc3 — the browser flow now has its own `AbortController` and identity guard mirroring the device flow: `cancelSignIn` actually aborts the callback server and the wait, a completion after cancel is ignored, and a stale flows timeout error is attributed only to itself instead of overwriting a newer attempt.

packages/extensions/extensions/openai-codex/index.ts
pcfreak30: Fixed in f84c5bc3 — the fallback now derives from the apps home-dir resolution (`AIDER_DESK_HOME_DIR`/`AIDER_DESK_DIR` mirroring constants.ts) and the README claim is corrected. Agree the proper fix is a data-dir accessor on `ExtensionContext` so extensions stop guessing paths — happy to contribute that follow-up to the extension framework separately.

packages/extensions/extensions/openai-codex/index.ts
pcfreak30: Fixed in f84c5bc3 — the model-call path now distinguishes "not authenticated" from a refresh failure: no tokens → the sign-in error; refresh failure → the actual cause ("Token refresh failed: {status/cause}" with the network reason surfaced). The silent quota variant still logs instead of throwing, so background UI stays quiet.

packages/extensions/extensions/openai-codex/index.ts
pcfreak30: Fixed in f84c5bc3 — polling is now conservative: pending is 403/404 (codex-rs contract, now referenced in a comment) plus 400; a 400 body with an error code other than `authorization_pending` fails fast with the code it got; `slow_down` backs the interval off by 5s; 408/429/5xx keep polling instead of aborting; anything else is a real failure. The usercode 404 message now also mentions "or not supported by the auth server" so the settings wild-goose-chase is limited.

wladimiiir (approved):
Thanks.

PR #934, opened 2026-09-05
feat(extensions): add skills context api

pcfreak30:
Adds a SkillsContext to the extensions api (list/find/resolve/create/update/writeSkillFile) backed by SkillManager, with shared authoring validation (name, frontmatter, size, path safety) and an atomic createSkill that allows project skills to shadow global ones.

The extension manager now caches one project skill manager per project, evicts it after event dispatch, and broadcasts skills updates so the ui refreshes after project-scoped writes.

The learn extension's skill-writer is refactored to delegate to the new context instead of reimplementing path resolution and file controls.

### DIFF
--- a/packages/common/src/extensions.ts
+++ b/packages/common/src/extensions.ts
@@ -1328,6 +1328,15 @@ export interface ProjectContext {
    * });
    */
   addDisposable(setup: () => (() => void | Promise<void>) | void): void;
+
+  /**
+   * Get the skills context for listing, finding, and writing AiderDesk skills.
+   * Backed by the same SkillManager used by the built-in skills tools, so skills created here
+   * are picked up by skill activation and the workspace skills panel.
+   * @returns SkillsContext instance (the project's SkillManager)
+   * @throws Error if no SkillManager is available
+   */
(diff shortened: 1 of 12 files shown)


packages/common/src/extensions.ts
```diff
@@ -1573,6 +1573,15 @@ export interface ExtensionContext {
...
   getMemoryContext(): MemoryContext;
 
+  /**
+   * Get the skills context for listing, finding, and writing AiderDesk skills.
+   * Backed by the same SkillManager used by the built-in skills tools, so skills created here
+   * are picked up by skill activation and the workspace skills panel.
+   * @returns SkillsContext instance (the project's SkillManager)
+   * @throws Error if no SkillManager is available (e.g., outside a project context)
+   */
+  getSkillContext(): SkillsContext;
```
wladimiiir: As `SkillsContext` can only be returned when project is available, this should be inside `ProjectContext` instead as `ExtensionContext` should return only app-level contexts.

wladimiiir (changes requested):
`SkillsContext` should be moved to `ProjectContext` as it's on project-level.

PR #939, opened 2026-09-06
feat: add copy task id to task menu

pcfreak30:
Adds a "Copy task id" item to the task sidebar "..." menu that writes the task's id to the clipboard. Uses the IPC clipboard API with a navigator.clipboard fallback, and shows a warning notification when copying fails.


wladimiiir:
I don't really want to make the menu bigger with actions that will get rarely used. What about just adding it into the Command Palette? That will be much faster to access if needed.

pcfreak30:
Closing — switching to a Command Palette approach.

PR #940, opened 2026-09-06
fix(extensions): speed up extension loading

pcfreak30:
Replaces per-file TypeScript `createProgram` parsing in the extension fetcher with `ts.createSourceFile`, eliminating full compiler-program construction for every extension entry file during gallery loading.


wladimiiir (approved):
Thanks.

PR #944, opened 2026-09-07
feat: add copy task id to command palette

pcfreak30:
Adds a "Copy task id" command to the command palette, writing the active task.id to the clipboard via the IPC clipboard API. Shows a notification on success and a warning when copying fails.


wladimiiir (approved):
Thanks.

PR #947, opened 2026-09-08
fix: restore welcome page on non-installed projects with BMAD mode switcher

777marvin:
### What changed

The v2 rewrite (69e8d4fa) gated `getUIComponents` on the BMAD installation, so non-installed projects showed no extension UI at all. That also removed the install entry point in BMAD mode.

The first version of this PR restored that entry point with a `welcome-page` component for the non-BMAD modes. That was wrong: any component at the `welcome-page` slot replaces the default welcome screen entirely, so agent and code mode lost the Code/Agent mode cards. The review screenshot in the thread showed that.

### The fix (follow-up commit)

`getUIComponents` now returns, for non-installed projects:

- no task context: no components
- BMAD mode: the BMAD welcome page with the install prompt (v1 behavior)
- other modes: a compact `bmad-mode-switcher` button at `task-top-bar-right`. That slot is additive and never replaces default UI, so the default welcome screen with the Code/Agent cards stays intact. The button switches via `TaskContext.updateTask({ currentMode: 'bmad' })`; the existing mode-change reload then swaps in the BMAD page.

Installed projects are unchanged: welcome page plus task actions in all modes.

### Verification

- Mode gating, the `switch-to-bmad` action and the no-task-context case are covered in `lib/index.test.ts`
- `tsc --noEmit` clean, test suite 186/186 (13 files) in the extension repo
- Manual check on a non-BMAD project after the fix; screenshot in the comment thread below

One correction to the earlier description of this PR: it said the UI change had been verified manually. That was not true when it was written, only the automated checks had run. The manual check happened after the fix, and the screenshot in the comment thread shows the result.


wladimiiir (changes requested):
image

It seems that the Welcome page is also replaced for other modes.

777marvin:
@wladimiiir thanks for the review and the screenshot, and sorry for the flawed first version.

The problem was my choice of slot: the switcher sat at `welcome-page`, and that placement replaces the default welcome screen in every mode. Agent and code mode therefore lost the Code/Agent cards.

The follow-up commit moves the button to `task-top-bar-right`, an additive slot. In agent and code mode the default welcome screen is intact again, BMAD mode still shows the BMAD page, and no components are returned when there is no task context.

One correction to the PR description: it said the change had been verified manually, but at that point only the automated checks had run. The manual check only happened after the rework. Screenshot of the result:

App screenshot after the rework

777marvin:
Addressed the review note: the unused \existsSync\ import is removed in 1261a64f. \	sc --noEmit\ clean, test suite 186/186.

wladimiiir:
We should show the default Welcome page in non-BMAD modes. I have made the change, so I am closing this now. 

See https://github.com/hotovo/aider-desk/commit/bee78d339edf959d4137178bd99f4aecff1bf07a for details.

PR #952, opened 2026-09-11
fix(tps): improve streaming and subagent rates

pcfreak30:
Adds real-time per-message TPS beside token and cost usage.

Scopes dashboard metrics by task hierarchy and shows active descendant subagent throughput separately from the parent rate.

Uses provider usage when available, streamed-token fallback otherwise, and prevents unrealistic early-window TPS spikes.

Adds focused regression coverage for streaming, fallback usage, grouped messages, and parent/subagent behavior.

Focused tests pass: 3 tests.


wladimiiir (approved):
Thanks

